PR #7, opened 2018-05-31
Adds option deserialization for inner value if Some(T)

tcr:
I kinda just banged this out, but this implements:

1. Serialize an inner value if Some(T), skip writing anything if None
2. Deserialize values into Some(T)

This should allow, e.g.:

```
struct Doc {
  #[serde(with = "serde_with::option::unwrap_or_skip")]
  value: Option<String>,
}
```

And have the value skipped if None, or written if Some.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,7 @@ pub mod chrono;
 #[cfg(feature = "json")]
 pub mod json;
 pub mod rust;
+pub mod option;
 
 /// Seperator for string-based collection de/serialization
 pub trait Separator {
--- a/src/option.rs
+++ b/src/option.rs
@@ -1,23 +1,23 @@
 //! De/Serialization of Option Types
 
-/// Serialize value if Some(..), skip if None
+/// Serialize inner value if Some(..). If none, serialize the unit struct ().
+/// When used with: (skip_serializing_if = "Option::is_none") and 
+/// serde(default), you can skip a value if it's None, or serialize its inner
+/// value if Some(T).
 pub mod unwrap_or_skip {
-    use serde::de::{DeserializeOwned, Deserializer, Error, Visitor};
-    use serde::ser::{self, Serialize, Serializer};
-    use serde_json;
-    use std::fmt;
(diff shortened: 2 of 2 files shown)


jonasbb:
Thanks for wanting to contribute.
If you just want to skip the serializing step if an Option is None, then it is easier to use `#[serde(skip_serializing_if = "Option::is_none")]`. This does not require any external crates. Deserializing a value which is not present will give you a None.

tcr:
@jonasbb Thanks for replying!

I put together [this gist as an example](https://gist.github.com/tcr/1d8dc5b2d324c8e6d69625e03587b639) of what I'm looking to do. (I also updated the branch.) When encoding with non-JSON outputs the `Some(...)` enum variant is encoded along with it, which is undesirable for what I want to accomplish. The output of this gist:

```
$ cargo run
   Compiling help v0.1.0 (file:///Users/trim/tcr/ok1)
    Finished dev [unoptimized + debuginfo] target(s) in 1.9 secs
     Running `target/debug/help`
serde_with
   a: "(always:\"hi\",)"
  a': "(always:\"hi\",)"
   b: "(always:\"hi\",optional:\"hello\",)"
  b': "(always:\"hi\",optional:\"hello\",)"
no deps
   a: "(always:\"hi\",)"
  a': "(always:\"hi\",)"
  b: "(always:\"hi\",optional:Some(\"hello\"),)"
  b': "(always:\"hi\",optional:Some(\"hello\"),)"
```

I wish it didn't also require `skip_serializing_if` and `default` in order to achieve these semantics, but essentially with it I can "inline" a field as though it were an optional field with no Some variant being exposed.

Let me know if I'm still overlooking something I can do with bare serde :)

jonasbb:
Using it with non-JSON serialization makes sense. Would you mind adding an example which also works as a doctest? That way there is a regression test and it is easier to see how to use it.
You can simply add `ron` as a dev-dependency.

Otherwise the PR looks good.

jonasbb:
I fixed your code and commited it.

PR #9, opened 2018-06-02
Macro to handle prefixed field names

dtolnay:
The [Challonge REST API] likes to use prefixes to group related fields. In simplified form, their JSON may resemble the following:

[Challonge REST API]: https://api.challonge.com/v1/documents/matches/show

```json
{
  "player1_name": "name1",
  "player1_votes": 1,
  "player2_name": "name2",
  "player2_votes": 2
}
```

In Rust we would ideally like to model this data as a pair of `Player` structs, rather than repeating the fields of `Player` for each prefix.

```rust
struct Match {
    player1: Player,
    player2: Player,
}

struct Player {
    name: String,
    votes: u64,
}
```

This `with_prefix!` macro produces an adapter that adds a prefix onto field names during serialization and trims away the prefix during deserialization. An implementation of the Challonge API would use
`with_prefix!` like this:

```rust
#[macro_use]
extern crate serde_derive;

#[macro_use]
extern crate serde_with;

extern crate serde_json;

struct Match {
    #[serde(flatten, with = "prefix_player1")]
    player1: Player,
    #[serde(flatten, with = "prefix_player2")]
    player2: Player,
}

struct Player {
    name: String,
    votes: u64,
}

with_prefix!(prefix_player1 "player1_");
with_prefix!(prefix_player2 "player2_");

fn main() {
    let m = Match {
        player1: Player {
            name: "name1".to_owned(),
            votes: 1,
        },
        player2: Player {
            name: "name2".to_owned(),
            votes: 2,
        },
    };

    let j = serde_json::to_string_pretty(&m).unwrap();
    println!("{}", j);
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,7 +63,9 @@
 
 #[cfg(feature = "chrono")]
 extern crate chrono as chrono_crate;
-extern crate serde;
+#[doc(hidden)]
+#[macro_use]
+pub extern crate serde;
 #[cfg(feature = "json")]
 extern crate serde_json;
 
@@ -72,6 +74,8 @@ pub mod chrono;
 #[cfg(feature = "json")]
 pub mod json;
 pub mod rust;
+#[doc(hidden)]
+pub mod with_prefix;
 
 /// Seperator for string-based collection de/serialization
 pub trait Separator {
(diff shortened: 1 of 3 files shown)


jonasbb:
Thank you, this looks good.
I hope you don't mind me taking this crate name. I will try to implement some more helpers suggested in the serde tracking issue.

bors r+

PR #24, opened 2018-11-22
Added `rust::seq_display_fromstr` to help de/serialize collections

katyo:
This helper uses `FromIterator`/`IntoIterator` traits for collection
itself and `Display`/`FromStr` for each element.

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -97,6 +97,127 @@ pub mod display_fromstr {
     }
 }
 
+/// De/Serialize sequences using [FromIterator] and [IntoIterator] implementation for it and [Display][] and [FromStr][] implementation for each element
+///
+/// This allows to serialize and deserialize collections with elements which can be represented as strings.
+///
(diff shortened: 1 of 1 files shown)


src/rust.rs
```diff
@@ -97,6 +97,127 @@ pub mod display_fromstr {
...
+            }
+        }
+
+        deserializer
+            .deserialize_seq(Helper(PhantomData))
+            .map(T::from_iter)
+    }
+
+    /// Serialize collection T using [IntoIterator] and [Display] for each element
+    pub fn serialize<S, T, I>(mimes: &T, serializer: S) -> Result<S::Ok, S::Error>
```
jonasbb: Could you please use a more general variable name here. `mime` looks like it is from a previous iteration of this code.

katyo: Ohh, it seems I forget to rename it when parametrizing.
Thanks.

jonasbb:
Thank you, looks good now.
I can release a new version later.

bors r+

Issue #32, opened 2019-02-07
Serialize a HashMap as a list of tuples


jonasbb:
https://github.com/serde-rs/serde/issues/1476

* [x] HashMap -> List of Tuples
* [x] BTreeMap -> List of Tuples
* [x] Vec<(T, T)> -> Map

dtolnay:
The Vec\<(K, V)\> ⟷ serde map is available in https://crates.io/crates/serde-tuple-vec-map but may make sense to reimplement in serde_with.

PR #33, opened 2019-02-16
Add documentation for HashMap and BTreeMap as tuple lists

jonasbb:
This addresses the first two wishes from #32.

@upsuper: Ping, since you originally proposed this for addition to serde.
It would be nice to know if it works for you, since I generalized your implementation.


upsuper:
I tested the hashmap one, and it works pretty well with my program, and I have [replaced my own code with a dependency to serde_with](https://github.com/rust-net-web/areweasyncyet.rs/commit/eb1e31c973f454246452523a8c601bf93dd26ee1)!

Issue #59, opened 2019-06-17
Trouble using #[skip_serializing_none]


Warfields:
I can't seem to get the attribute to work.

```rust
use serde::{Serialize, Deserialize};
use serde_with;

// Makes all struct members public
macro_rules! pub_struct {
    ($name:ident {$($field:ident: $t:ty,)*}) => {
        #[derive(Deserialize, Serialize, Debug, Default)] // ewww
        pub struct $name {
            $(pub $field: $t),*
        }
    }
}

#[derive(Deserialize, Serialize, Debug)]
pub struct QueryList<T>{
    pub query: Vec<T>,
}

pub_struct!(QuerySimple{
    endpoint: String,
});

pub_struct!(Criteria {
    op: String,
    field: String,
    value: String,
    criteria: Vec<Criteria>,
});

pub_struct!(Reference{
    reference: String,
    references: Vec<Reference>,
});

#[skip_serializing_none]
#[derive(Serialize, Default)]
pub struct Query {
    endpoint: String,
    id_int64: Option<i64>,
    id_string: String,
    criteria: Criteria,
    references: Vec<Reference>,
}
```

jonasbb:
You cannot use a macro, without importing it first. In your code above, that is missing.
If you change `use serde_with;` to `use serde_with::skip_serializing_none;` or you change `#[skip_serializing_none]` to `#[serde_with::skip_serializing_none]`, then it compiles.

Do you have any other problems with the code expect from getting it to compile?

Warfields:
Nope! Thanks for the help. I was expecting that it had the `#[macro_export]` annotation.

Issue #49, opened 2019-04-02
Documentation for inline proc-macros


jonasbb:
Normally re-exported items can be documented directly, without the re-export section in rustdoc.
This is not possible for proc-macro functions due to a limitation in rust, tracked here: https://github.com/rust-lang/rust/issues/58696.

What alternatives are there for documentation?

jonasbb:
This is solved since a while.

PR #60, opened 2019-06-18
Updated documentation

Warfields:
Which crate to use was not obvious since serde_with_macros appears as its own crate on docs.rs

### DIFF
--- a/serde_with_macros/src/lib.rs
+++ b/serde_with_macros/src/lib.rs
@@ -66,6 +66,8 @@ use syn::{
 /// #
 /// # use serde::Serialize;
 /// # use serde_with_macros::skip_serializing_none;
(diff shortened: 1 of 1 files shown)


Warfields:
To prevent issues like #59

jonasbb:
Hi @Warfields, thanks for the effort. Unfortunately, I don't think it is possible to do exactly what you were planning here. The `serde_with` crate cannot be used from within `serde_with_macros`, as otherwise this would create a circular dependency between the two crates.

I would prefer to use the `#[doc(inline)]` attribute on the proc_macro, but this doesn't work, since there is a rustc limitation right now (https://github.com/jonasbb/serde_with/issues/49).
I will improve the documentation with adding some hints, that everything from `serde_with_macros` is re-exported in `serde_with`.

PR #67, opened 2019-08-25
Improve CI pipeline

jonasbb:
* Run cargo audit
* Use tarpaulin also on stable Rust versions
* Bump minimal version to 1.31.1 to use Edition 2018


jonasbb:
bors try

PR #69, opened 2019-08-27
Add trybuild tests to check the error messages created by the macros

### LINKED ISSUE #50: Add compiletest to test the error messages of proc_macros in serde_with_macros
The [compiletest](https://github.com/laumann/compiletest-rs) crate can be used to test for error messages of proc_macros.

Currently it has a problem with running cargo check and cargo test after each other. Since we test also using clippy, this prevents compiletest from working correctly. The issue is tracked at https://github.com/laumann/compiletest-rs/issues/101.

An initial test setup using compiletest is in the [compiletest branch](https://github.com/jonasbb/serde_with/tree/compiletest).

jonasbb:
Use trybuild as an improvement over compiletest as the test framework
for the UI tests.
Older Rust versions (<1.35.0) have different error messages, so the new
UI tests fail on them, therefore gate them on the Rust version.

Closes #50

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,6 +91,7 @@ after_success:
   - |
     if [[ "$TRAVIS_RUST_VERSION" == stable ]] || [[ "$TRAVIS_RUST_VERSION" == nightly ]]
     then
+      export TARPAULIN=1
       cargo tarpaulin --out Xml --all --all-features
       bash <(curl -s https://codecov.io/bash)
       echo "CodeCov Uploading of data was successful!"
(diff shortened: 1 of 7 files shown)


jonasbb:
Thank you for the hint with the attribute. Yesterday I looked for crates doing rustc version checking but only found [`rustc_version`](https://docs.rs/rustc_version/0.2.3/rustc_version/) and [`version_check`](https://docs.rs/version_check/0.9.1/version_check/) but not your `rustversion` crate.

PR #73, opened 2019-10-10
Check how the using doctests for code coverage changes the results

jonasbb:


jonasbb:
Tarpaulin is now able to use doctests and show coverage information.
However, the line information for the doc-block part are wrong and mean that other unrelated lines on the top of the file will be marked as covered. More information here: https://github.com/xd009642/tarpaulin/issues/215#issuecomment-538357957

PR #101, opened 2020-05-10
Improve performance of applying prefix

peterjoel:
`format!("{}{}", a, b)` is a very slow method of concatenating strings. See benchmark: https://gist.github.com/peterjoel/97431103893703e81926ae9bd8f768c5#file-results-txt-L2

### DIFF
--- a/src/with_prefix.rs
+++ b/src/with_prefix.rs
@@ -390,8 +390,10 @@ where
     where
         T: ?Sized + Serialize,
     {
-        self.delegate
-            .serialize_entry(&format!("{}{}", self.prefix, key), value)
+        let mut prefixed_key = String::with_capacity(self.prefix.len() + key.len());
+        prefixed_key.push_str(self.prefix);
+        prefixed_key.push_str(key);
+        self.delegate.serialize_entry(&prefixed_key, value)
     }
 
     fn end(self) -> Result<Self::Ok, Self::Error> {


src/with_prefix.rs
```diff
@@ -390,8 +390,10 @@ where
     where
         T: ?Sized + Serialize,
     {
-        self.delegate
-            .serialize_entry(&format!("{}{}", self.prefix, key), value)
+        let mut prefixed_key = String::with_capacity(self.prefix.len() + key.len());
+        prefixed_key.push_str(self.prefix);
+        prefixed_key.push_str(key);
+        self.delegate.serialize_entry(&prefixed_key, value)
```
dtolnay, line 396: Ideally this would serialize the key using https://docs.rs/serde/1.0.108/serde/trait.Serializer.html#method.collect_str to avoid allocating at all in most formats.
https://github.com/serde-rs/json/blob/eb383c08b234ce1f134838155f784bedf35a7aea/src/ser.rs#L455

peterjoel: Yep, I'll change that.

It's a shame that serializers that can't (or just don't) implement a non-allocating `collect_str` will perform even worse by changing this.

jonasbb:
Thank you, this change looks good.

@bors r+

jonasbb:
Let's give this PR another try with an updated CI configuration.

bors r+

PR #145, opened 2020-08-08
Test new codecov configuration

jonasbb:


jonasbb:
codecov does not seem to use the config from the PR. Let's see the result after merging it.

bors r+

PR #186, opened 2020-10-04
namespace Result to std::result

adwhit:
I often like to alias `type Result<T> = std::result::Result<T, MyError>`.

Unfortunately this doesn't work with the proc-macros as they assume `Result` == `std::result::Result`. The fix is to fully namespace them.

### DIFF
--- a/serde_with_macros/src/lib.rs
+++ b/serde_with_macros/src/lib.rs
@@ -714,7 +714,7 @@ fn deserialize_fromstr(item: TokenStream) -> Result<proc_macro2::TokenStream, Er
             Self: std::str::FromStr,
             <Self as FromStr>::Err: std::fmt::Display,
         {
-            fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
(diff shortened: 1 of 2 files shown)


jonasbb:
Thanks for the contribution. This is an oversight of me. Do you mind adding a test for this? You can take inspiration from https://github.com/jonasbb/serde_with/blob/9b56c3ab1d09a5d09c88a924fc9ba1b09b8a3a6b/serde_with_macros/tests/serialize_display.rs and https://github.com/jonasbb/serde_with/blob/9b56c3ab1d09a5d09c88a924fc9ba1b09b8a3a6b/serde_with_macros/tests/deserialize_fromstr.rs

adwhit:
Have added a test, let me know if it's what you had in mind. Also namespaced the `FromStr` trait so you don't have to import it.

jonasbb:
Yes that it exactly what I had in mind. Thank you so much.
I will prepare a patch release update after this merges.

bors r+

PR #188, opened 2020-10-05
deserialize macro implements visit_bytes

adwhit:
I hit a problem when using  the `DeserializeFromStr` macro along with the `csv` crate, namely that `csv` tries to call `visit_bytes` to deserialize each column, and the macro does not implement this method. This PR adds it.

I have added a slightly convoluted test case which creates a `Deserializer` which can only handle bytes. This was necessary as the `serde_json` crate used elsewhere is 'too clever' for this test.

### DIFF
--- a/serde_with_macros/src/lib.rs
+++ b/serde_with_macros/src/lib.rs
@@ -737,6 +737,14 @@ fn deserialize_fromstr(item: TokenStream) -> Result<proc_macro2::TokenStream, Er
                     {
                         value.parse::<Self::Value>().map_err(serde::de::Error::custom)
                     }
+
+                    fn visit_bytes<E>(self, value: &[u8]) -> std::result::Result<Self::Value, E>
+                    where
+                        E: serde::de::Error,
+                    {
+                        let utf8 = std::str::from_utf8(value).map_err(serde::de::Error::custom)?;
+                        self.visit_str(utf8)
+                    }
                 }
 
                 deserializer.deserialize_str(Helper(std::marker::PhantomData))
--- a/serde_with_macros/tests/deserialize_fromstr.rs
+++ b/serde_with_macros/tests/deserialize_fromstr.rs
@@ -42,6 +42,46 @@ fn test_deserialize_fromstr() {
 }
 
 #[test]
+fn test_deserialize_from_bytes() {
+    use serde::de::{value::Error, Deserialize, Deserializer, Visitor};
+
+    // Unfortunately serde_json is too clever (i.e. handles bytes gracefully)
+    // so instead create a custom deserializer which can only deserialize bytes.
+    // All other deserialize_* fns are forwarded to deserialize_bytes
+    struct ByteDeserializer(&'static [u8]);
+
+    impl<'de> Deserializer<'de> for ByteDeserializer {
+        type Error = Error;
+
+        fn deserialize_any<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+        where
+            V: Visitor<'de>,
+        {
+            self.deserialize_bytes(visitor)
+        }
+
+        fn deserialize_bytes<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+        where
+            V: Visitor<'de>,
+        {
+            visitor.visit_bytes(self.0)
+        }
+
+        serde::forward_to_deserialize_any! {
+            bool i8 i16 i32 i64 i128 u8 u16 u32 u64 u128 f32 f64 char str string
+            byte_buf option unit unit_struct newtype_struct seq tuple
+            tuple_struct map struct enum identifier ignored_any
+        }
+    }
+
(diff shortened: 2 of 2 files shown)


jonasbb:
Hi, thank you for the next contribution :) It seems reasonable to treat bytes as a string here. So I will merge this tomorrow.

Do you have an example how this breaks usage with the `csv` crate? I tried some simple cases with the csv crate but all of them work with the current version. If you don't have an example that's fine too. The test looks good


<details>
<summary>My csv + serde_with test script</summary>

```rust
#!/usr/bin/env scriptisto

// scriptisto-begin
// script_src: src/main.rs
// build_cmd: cargo build --release && strip ./target/release/script
// target_bin: ./target/release/script
// files:
//  - path: Cargo.toml
//    content: |
//     package = { name = "script", version = "0.1.0", edition = "2018"}
//     [dependencies]
//     csv = "1.1.3"
//     serde = {version = "1.0.116", features = ["derive"]}
//     serde_with = "1.5.0"
// scriptisto-end

use serde::*;
use serde_with::*;

#[derive(Debug, Deserialize)]
struct Record {
    city: String,
    a: A,
    num: u32,
}

// Type taken from the test case.
#[derive(Debug, PartialEq, DeserializeFromStr)]
struct A {
    a: u32,
    b: bool,
}

impl std::str::FromStr for A {
    type Err = String;

    /// Parse a value like `123<>true`
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let mut parts = s.split("<>");
        let number = parts
            .next()
            .ok_or_else(|| "Missing first value".to_string())?
            .parse()
            .map_err(|err: std::num::ParseIntError| err.to_string())?;
        let bool = parts
            .next()
            .ok_or_else(|| "Missing second value".to_string())?
            .parse()
            .map_err(|err: std::str::ParseBoolError| err.to_string())?;
        Ok(Self { a: number, b: bool })
    }
}

fn main() -> Result<(), Box<dyn std::error::Error>>{
    let data = r#"
city,a,num
Boston,"159<>true",4628910
Concord,555<>false,42695
"#;

    let mut rdr = csv::Reader::from_reader(data.as_bytes());
    for result in rdr.deserialize() {
        let record: Record = result?;
        println!("{:?}", record);
    }
    println!("-----");
    let mut rdr = csv::Reader::from_reader(data.as_bytes());
    for result in rdr.into_deserialize() {
        let record: Record = result?;
        println!("{:?}", record);
    }
    println!("-----");
    let mut rdr = csv::Reader::from_reader(data.as_bytes());
    for result in rdr.into_byte_records() {
        let record: Record = result?.deserialize(None)?;
        println!("{:?}", record);
    }
    println!("-----");
    let record: Record =
        csv::ByteRecord::from(vec!["Boston", "159<>true", "4628910"]).deserialize(None)?;
    println!("{:?}", record);
    println!("-----");
    let record: Record =
        csv::StringRecord::from(vec!["Boston", "159<>true", "4628910"]).deserialize(None)?;
    println!("{:?}", record);

    Ok(())
}
```

which prints

```
Record { city: "Boston", a: A { a: 159, b: true }, num: 4628910 }
Record { city: "Concord", a: A { a: 555, b: false }, num: 42695 }
-----
Record { city: "Boston", a: A { a: 159, b: true }, num: 4628910 }
Record { city: "Concord", a: A { a: 555, b: false }, num: 42695 }
-----
Record { city: "Boston", a: A { a: 159, b: true }, num: 4628910 }
Record { city: "Concord", a: A { a: 555, b: false }, num: 42695 }
-----
Record { city: "Boston", a: A { a: 159, b: true }, num: 4628910 }
-----
Record { city: "Boston", a: A { a: 159, b: true }, num: 4628910 }
```

</details>

adwhit:
Hi, thanks for your comment.

Here is an example that fails at the moment but succeeds with the patch. I think the problem lies with how the header is deserialized. I may have a look into the `csv` crate to see if some improvement can be made there too.

```
use serde_with::DeserializeFromStr;
use std::collections::HashMap;

#[derive(Debug, DeserializeFromStr, PartialEq, Eq, Hash)]
struct Data {
    left: String,
    right: String
}

impl std::str::FromStr for Data {
    type Err = String;
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let mut iter = s.split('-');
        match (iter.next(), iter.next(), iter.next()) {
            (Some(left), Some(right), None) => {
                Ok(Data {
                    left: left.into(),
                    right: right.into(),
                })
            }
            _other => Err(format!("Could not parse {}", s))
        }
    }
}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let data = "
a-b,c-d
hello,world
";
    let mut rdr = csv::Reader::from_reader(std::io::Cursor::new(data));
    let res = rdr.deserialize().next().unwrap();
    let r: HashMap<Data, String> = res?;
    println!("{:?}", r);
    Ok(())
}
```

jonasbb:
So it seems to be specific to how the header is deserialized. Interesting.

bors r+

Issue #201, opened 2020-10-30
serde_as fails for externally defined types


sergiimk:
Hi,

I'm somewhat new to the Rust language and having a rough time with `serde` and externally defined types that are nested into other data structures.

As shown in example below I have a struct `A` defined in another crate which I'd like to serialize without converting to another type, so I'm using the [remote derive](https://serde.rs/remote-derive.html) method. 

Using `#[serde(with = "ADef")]` worked fine for the plain type `A` but composite types like `Option<A>` were more challenging and lead me to your crate.

With some hackery I was able to get `#[serde_as(as = "Option<ADef>")]` working, but trying to serialize `BTreeMap<String, A>` has finally defeated me and I'd really appreciate some tips.

```rust
use serde::{Deserialize, Deserializer, Serialize, Serializer};
use serde_with::{serde_as, DeserializeAs, SerializeAs};
use std::collections::BTreeMap;

mod external {
    #[derive(Debug, Clone)]
    pub struct A {
        pub x: String,
    }
}

use external::A;

#[serde(remote = "A")]
#[derive(Debug, Clone, Serialize, Deserialize)]
struct ADef {
    pub x: String,
}

#[serde_as]
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct Foo {
    #[serde(with = "ADef")]
    pub inst: A,
    #[serde_as(as = "Option<ADef>")]
    pub opt: Option<A>,

    // Can't get this working at all
    // #[serde_as(as = "BTreeMap<String, ADef>")]
    // pub map: BTreeMap<String, A>,

    // Confusingly enough defining identity transform like this doesn't compile either
    // #[serde_as(as = "BTreeMap<String, String>")]
    // pub map2: BTreeMap<String, String>,
}

// Required for `as = "Option<ADef>"` to work :(
impl SerializeAs<A> for ADef {
    fn serialize_as<S>(source: &A, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        #[derive(Serialize)]
        struct Helper<'a>(#[serde(with = "ADef")] &'a A);
        Helper(source).serialize(serializer)
    }
}
impl<'de> DeserializeAs<'de, A> for ADef {
    fn deserialize_as<D>(deserializer: D) -> Result<A, D::Error>
    where
        D: Deserializer<'de>,
    {
        #[derive(Deserialize)]
        struct Helper(#[serde(with = "ADef")] A);
[...]

PR #226, opened 2020-12-10
Better document how to use `SerializeAs` with types from other crates

### LINKED ISSUE #225: Using #[serde_as] with custom (de-)serialize functions
I have implemented custom (de-)serialize functions for a remote type, e.g. `Py<T>` (doesn't really matter what exactly it is). With `serde`, I can simply use these custom functions via either `#[serde(with = "custom_serde")]` (where `custom_serde` is a module containing the (de-)serialize function) or `#[serde(serialize_with = "custom_serde::serialize")]` and `#[serde(deserialize_with = "custom_serde::deserialize")]`.

However, I also have structs that contain e.g. `Vec<Py<Inner>>`, which cannot be (de-)serialized with this approach. To my understanding, this is exactly the conundrum that `serde_with` is supposed to solve. From my research, however, I only found ways to specify a type to serialize with, i.e., I would need to implement the (De-)Serialize traits for `Py<T>` directly rather than forward to a custom implementation via annotation.

Are free-standing (de-)serializers not yet supported or am I missing something? In any case I think the documentation could use some work here - `#[serde_as]` is meant to replace `#[serde(with = "...")]` yet doesn't seem to support its basic features such as declaring a module which contains (de-)serialize functions.

I don't think I can implement (De-)Serialize for `Py<T>` due to the orphan rule and the workaround that serde supports would rely on the internal representation of `Py<T>`, which I don't care for and thus would rather not depend on. Please feel free to correct me if there's an easy way I'm missing.

jonasbb:
1. Describe how to do this with an existing module compatible with the
   #[serde(with = "...")] annotation.
   Based on the discussion in #225.
2. Based on serde's remote derive. This has a similar use case.

Closes #225

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Some common use cases are:
 * Deserialize a comma separated list like `#hash,#tags,#are,#great` into a `Vec<String>`.
      Check the documentation for [`serde_with::rust::StringWithSeparator::<CommaSeparator>`][StringWithSeparator].
 
-Check out the [**user guide**][user guide] to find out more tips and tricks about this crate.
+**Check out the [user guide][user guide] to find out more tips and tricks about this crate.**
 
 ## Use `serde_with` in your Project
 
--- a/src/guide.md
+++ b/src/guide.md
@@ -6,61 +6,7 @@ For example, you can serialize [a map as a sequence of tuples][crate::guide::ser
 
 The crate offers four types of functionality.
 
-## 1. Integration with serde's with-annotation
-
-[serde's with-annotation][with-annotation] allows to specify a different serialization or deserialization function for a field.
-It is usefull to adapt the serialization of existing types to the requirements of a protocol.
-Most modules in this crate can be used together with the with-annotation.
-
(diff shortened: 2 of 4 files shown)


jonasbb:
@Lehona Do you think this description would have been usefull for you? Based on the discussion in #225.
[Using `#[serde_as]` on types without `SerializeAs` and `Serialize` implementations](https://github.com/jonasbb/serde_with/blob/2ab93e3ca624cc81f9c2aca2f5bd23c0ad088f15/src/guide/serde_as.md#using-serde_as-on-types-without-serializeas-and-serialize-implementations)

@sergiimk I have seen you use `#[serde_as]` with the remote derive feature, so you might be interested in this section (like #201)
[Using `#[serde_as]` with serde's remote derives](https://github.com/jonasbb/serde_with/blob/2ab93e3ca624cc81f9c2aca2f5bd23c0ad088f15/src/guide/serde_as.md#using-serde_as-with-serdes-remote-derives)

Lehona:
Yes, that's great!

jonasbb:
Thank you for taking a look.

Issue #250, opened 2021-01-28
Provide a deserialize_ignore_any function


lovasoa:
Hello !
  Would you be interested in providing the following function : 

```rust
fn deserialize_ignore_any<'de, D: Deserializer<'de>, T:Default>(deserializer: D) -> Result<T, D::Error> {
    serde::de::IgnoredAny::deserialize(deserializer).map(|_| T::default())
}
```

It is useful in combination with `#[serde(deserialize_with = "...")` when a field should accept any value without actually storing  its contents. 

In particular, it can be used to work around the fact that `#[serde(other)]` doesn't work with non-unit enum variants, with for instance : 

```rust
#[derive(Deserialize)]
enum Item {
    Foo, Bar, Baz,
    #[serde(other, deserialize_with = "deserialize_ignore_any")]
    Other,
}
```

jonasbb:
Hi, thanks for the interest in this project. The function fits the scope of this project and could be added.

However, it seems your example does not what you intend it to do:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bec97a5a7ea17fc21ba46af3e80d1453

Do you have another use-case for this function or an alternative implementation?

lovasoa:
The function does work as expected. I'm not sure I understand what you are trying to do in the playground example. If you are trying to deserialize a list of strings, you don't need deserialize_ignore_any at all. It is helpful in cases like this, for example: https://github.com/RReverser/serde-xml-rs/issues/55#issuecomment-768902802

jonasbb:
`#[serde(other)]` declares an enum variant as fallback. `"Ba"` does not deserialize to any enum variant so in the playground it falls back to `Other`. However, if I add the `deserialize_with` attribute I get this error while deserializing: `Err(Error("invalid type: unit variant, expected newtype variant", line: 0, column: 0))`. The code deserializes the correct enum variant if you change `"Ba"` to `"Foo"`, `"Bar"`, or `"Baz"`.

I will check the serde-xml link later.

lovasoa:
Ok, I understand what you were trying to do. Indeed, deserialize_ignore_any does not work in that case, because `deserialize_with` hints the serde that you are trying to deserialize the contents of the enum variant, and the string **is** the enum variant, but does not have any contents. If you want an example where using  deserialize_ignore_any may be useful in conjunction with `other` in serde_json, see for instance: 

```rust
use serde::*;

fn main() {
[...]

PR #251, opened 2021-01-28
add deserialize_ignore_any

lovasoa:
Fixes https://github.com/jonasbb/serde_with/issues/250

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -1806,16 +1806,16 @@ pub mod default_on_null {
 /// Deserialize any value, ignore it, and return the default value for the type being deserialized.
 /// It is useful for instance to create an enum with a catch-all variant
 /// that will accept any incoming data.
-/// 
+///
 ///
 /// ## Example: deserializing a heterogeneous collection of xml nodes
-/// 
+///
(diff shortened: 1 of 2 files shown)


src/rust.rs
```diff
@@ -1802,3 +1802,23 @@ pub mod default_on_null {
         value.serialize(serializer)
     }
 }
+
+/// Deserialize any value, ignore it, and return the default value for the type being deserialized
```
jonasbb, line 1806: Could you please extend the description with a note that this only works for enums with data.
And maybe also the part about simulating an other-like variant in untagged enums. Basically the info in my last comment in #250.

jonasbb (changes requested):
Thanks for the contribution. Could you please extend the description a bit.

If you want a test case would be nice to have but I am happy to merge it already and add the test case later.

src/rust.rs
lovasoa: I added a description giving more context about when and why the function should be used.

jonasbb:
Thanks for updating the description. Once CI passes you can merge the PR by writing `bors r+` in a comment.

bors delegate=lovasoa

PR #262, opened 2021-02-06
Change processing of `#[serde_as(...)]` attributes on fields.

jonasbb:
The attributes will no longer be stripped during proc-macro processing.
Instead, a private derive macro is applied to the struct/enum which captures them and makes them inert, thus allowing compilation.

This should have no effect on the generated code and on the runtime behavior.
It eases integration of third-party crates with `serde_with`, since they can now process the `#[serde_as(...)]` field attributes reliably.
Before this was impossible for derive macros and lead to akward ordering constraints on the attribute macros.

Previous discussion in: https://github.com/jonasbb/serde_with/discussions/260

@Lehona Could you please check if this solves your issue?

### DIFF
--- a/serde_with_macros/CHANGELOG.md
+++ b/serde_with_macros/CHANGELOG.md
@@ -12,6 +12,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Improve error messages when `#[serde_as(..)]` is misused as a field attribute.
     Thanks to @Lehona for reporting the bug in #233.
 * Internal cleanup for assembling and parsing attributes during `serde_as` processing.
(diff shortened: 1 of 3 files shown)


jonasbb:
@Lehona It would be great if you could take a look at this and if this fixes your proc-macro ordering problem.

Lehona:
I will try to allocate some time for that at work, however the issue has been sent to the ~shadowrealm~ backlog for now. I hope to be able to set aside some time on the weekend.

jonasbb:
@Lehona Thanks for the info. Then I am just going to merge this PR and release a new version. This should work, after all the attributes are no longer deleted. If you notice it doesn't quite solve the ordering problem just write another message and I will check that.

bors r+

PR #268, opened 2021-02-16
Fix compilation with serde_as if there is a field without annotation

### LINKED ISSUE #267: v1.6.3 (v1.4.0 macros) Breaks Un-annotated Fields
I have a struct like this:

``` rust
#[serde_with::serde_as]
#[derive(Serialize, Deserialize, Debug, PartialEq)]
pub struct Thing {
    pub id: u8,

    #[serde_as(as = "Vec<serde_with::hex::Hex>")]
    pub data: Vec<Vec<u8>>,
}
```

It worked well until today, presumably with the release of v1.6.3, when I started getting this error:

```
error: An empty `serde_as` attribute on a field has no effect. You are missing an `as`, `serialize_as`, or `deserialize_as` parameter.
  --> src/thing:14:5
   |
14 |     pub id: u8,
   |     ^^^
```

I did not previously have to annotate the `id` field. So I tried adding an annotation:

``` rust
#[serde_with::serde_as]
#[derive(Serialize, Deserialize, Debug, PartialEq)]
pub struct Thing {
    #[serde_as(as = "u8")]
    pub id: u8,

    #[serde_as(as = "Vec<serde_with::hex::Hex>")]
    pub data: Vec<Vec<u8>>,
}
```

But now it says:

```
error[E0277]: the trait bound `u8: SerializeAs<_>` is not satisfied
   --> src/thing:11:1
    |
11  | / #[serde_with::serde_as]
12  | | #[derive(Serialize, Deserialize, Debug, PartialEq)]
    | |__________________^ the trait `SerializeAs<_>` is not implemented for `u8`
    | 
```

So how do I get the default serialization behavior back again?

jonasbb:
This broke in 1.4.0 due to removing the check if there are any `serde_as` attributes.
Add a test case for a struct with an un-annotated field.
    
Fixes #267

Bump versions to release bugfix.

bors r+

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.6.4]
+
+### Fixed
+
+* Fix compiling when having a struct field without the `serde_as` annotation by updating `serde_with_macros`.
+    This broke in 1.4.0 of `serde_with_macros`. [#267](https://github.com/jonasbb/serde_with/issues/267)
+
 ## [1.6.3]
 
 ### Changed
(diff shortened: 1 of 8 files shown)


imp:
Thanks for a quick fix. Is it possible to publish to crates.io?

PR #271, opened 2021-03-07
Big Array Support

jonasbb:


jonasbb:
https://github.com/est31/serde-big-array/issues/6

PR #295, opened 2021-03-31
Fix serde_as link target in readme

Ch00k:


jonasbb:
Thanks, that seems like a better link target.

bors r+

PR #301, opened 2021-04-18
Add array deserialization support to `Hex`

### LINKED ISSUE #300: All types of Bytes should be available in hex
I have a `[u8, N]` and want to deserialize it in the hex format. Base64 would also be cool but also way more complex.

TheAlgorythm:
I did what I thought you described on #300.
Please let me know if there are any issues.

Closes #300

### DIFF
--- a/src/hex.rs
+++ b/src/hex.rs
@@ -67,6 +67,45 @@ use std::marker::PhantomData;
 ///     BytesUppercase(vec![0x00, 0xaa, 0xbc, 0x99, 0xff]),
 ///     serde_json::from_value(json!("00aAbc99FF")).unwrap()
 /// );
+///
+/// /////////////////////////////////////
+/// // Arrays are supported in Rust 1.48+
+///
+/// # #[rustversion::since(1.48)]
+/// # fn test_array() {
+/// #[serde_as]
+/// # #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


jonasbb:
Thanks for the contribution. I had something like this in mind
```rust
impl<const N: usize> SerializeAs<[u8; N]> for HexArray<Lowercase> {
    fn serialize_as<S>(source: &[u8; N], serializer: S) -> Result<S::Ok, S::Error>
```
however, I think your solution is more generic, i.e., applies to more types. Since the is a blanket implementation of `TryFrom` for `Into`, I think the new type is not necessary. Instead replacing the `DeserializeAs` implementation with your code works.

The test error you see is because the `TryFrom<Vec<T>>` implementation for arrays is only available in Rust 1.48+. You need to make sure the code is only executed on newer Rust versions.

jonasbb:
I updated your pull request. I hope you don't mind.

TheAlgorythm:
Yes, that is the better solution. I did upgrade the feature-flag but forgot to push it. I tried a separate implementation for `TryFrom` but that didn't work as of `conflicting implementation`s. That should have hinted me to delete the `From`implementation.
The only problem I can imagine of is that the [blanket implementation](https://doc.rust-lang.org/std/convert/trait.TryFrom.html#impl-TryFrom%3CU%3E) is stable since `1.34`.

Issue #309, opened 2021-04-28
Macros for serializing a type as another type


Boscop:
I've been using these macros for a long time, to serialize a type as a different repr type.
They make it easy to serialize types from other crates that aren't `Serialize`/`Deserialize`.
I think it makes sense to include these macros in this crate, what do you think?

```rust
/// serializes a type as a different repr type that it can be converted Into and From
#[macro_export]
macro_rules! serde_as {
	($m:ident, $t:ty, $r:ty) => {
		pub mod $m {
			use super::*;
			use serde::{Deserialize, Deserializer, Serialize, Serializer};

			pub fn serialize<S: Serializer>(x: &$t, serializer: S) -> Result<S::Ok, S::Error> {
				let y: $r = (*x).into();
				y.serialize(serializer)
			}

			pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<$t, D::Error> {
				let y: $r = Deserialize::deserialize(deserializer)?;
				Ok(y.into())
			}
		}
	};
}

/// serializes a type as a different repr type using the given conversion functions
#[macro_export]
macro_rules! serde_conv {
	($m:ident, $t:ty, $ser:expr, $de:expr) => {
		pub mod $m {
			use super::*;
			use serde::{Deserialize, Deserializer, Serialize, Serializer};

			pub fn serialize<S: Serializer>(x: &$t, serializer: S) -> Result<S::Ok, S::Error> {
				let y = $ser(*x);
				y.serialize(serializer)
			}

			pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<$t, D::Error> {
				let y = Deserialize::deserialize(deserializer)?;
				Ok($de(y))
			}
		}
	};
}
```

Example usage:
```rust
serde_conv!(serde_timestamp, DOMTimeStamp, |x: DOMTimeStamp| x.0 as f64, |x: f64| DOMTimeStamp(x as _));
```
And then
```rust
#[serde(with = "serde_timestamp")] 
timestamp: DOMTimeStamp
```

Or:
```rust
serde_conv!(bool_str, bool, |x: &bool| x.to_string(), |x: String| x.parse::<bool>());

serde_conv!(flags_hashmap_bool_str, HashMap<Flag, bool>,
	|x: &HashMap<Flag, bool>| {
		x.iter().map(|(k, v)| (*k, v.to_string()))
			.collect::<HashMap<Flag, String>>()
	},
[...]

PR #310, opened 2021-04-29
Initial serde_conv implementation to define new serialization types with less boilerplate

jonasbb:
The design of serde_conv is initially by @Boscop in #309.
The addition of the `SerializeAs`/`DeserializeAs` traits is by me.

Part of #309


jonasbb:
I think this is basically done. It has some changes compared to #309. It has gained the support for `#[serde_as]`. The macro also gained an optional visibility specifier. I ran into a warning that a private type was in a public API. This can be avoided by not providing any visibility.

@Boscop I would be grateful if you could test this PR and #198. If everything works or if anything is missing for #309.

bors r+

PR #320, opened 2021-05-23
`serde_conv`: Allow `clippy::ptr_arg`

mkroening:
With `String` as `t`, `&String` would trigger [`clippy::ptr_arg`](https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg).

What do you think?


jonasbb:
Thank you. Yeah it makes sense to suppress the lints here. I wonder if there are other ones which might crop up. I guess they can be dealt with then.

bors r+

PR #325, opened 2021-06-14
Bytes const size borrowed arrays

matix2267:
Currently `Bytes` (added in #277) supports some const size array types (e.g. `[u8; N]` and `Box<[u8; N]>`)
but not others (e.g. `&[u8; N]` and `Cow<'_, [u8; N]>`).

This pull request implements `Bytes` support for `&[u8; N]` and `Cow<'_, [u8; N]>`
This bring parity between dynamic and const sized types.

### Before:

| Sizing  | Owned          | Borrowed      | Cow                   | Box               |
| ------- | -------------- | ------------- | --------------------- | ----------------- |
| Const   | ✅ `[u8; N]`   | ❌ `&[u8; N]` | ❌ `Cow<'_, [u8; N]>` | ✅ `Box<[u8; N]>` |
| Dynamic | ✅ `Vec<u8>`\* | ✅ `&[u8]`    | ✅ `Cow<'_, [u8]>`    | ✅ `Box<[u8]>`    |

### After:

| Sizing  | Owned          | Borrowed      | Cow                   | Box               |
| ------- | -------------- | ------------- | --------------------- | ----------------- |
| Const   | ✅ `[u8; N]`   | ✅ `&[u8; N]` | ✅ `Cow<'_, [u8; N]>` | ✅ `Box<[u8; N]>` |
| Dynamic | ✅ `Vec<u8>`\* | ✅ `&[u8]`    | ✅ `Cow<'_, [u8]>`    | ✅ `Box<[u8]>`    |

I've also added a separate commit with an example of using fully borrowed types, but that example only does serialization
because RON doesn't support 0-copy deserialization for byte arrays (as it needs to decode them from base64).
Because of that I'm not sure if this example is useful so feel free to merge only the main commit.

<sub>\* Technically `[u8]` corresponds to `[u8; N]` but `[u8]` is an un`Sized` type and cannot be
directly used in structs.</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+* The `Bytes` type now supports borrowed and Cow arrays of fixed size (requires Rust 1.51+)
+
+    ```rust
+    #[serde_as(as = "Bytes")]
+    #[serde(borrow)]
+    borrowed_array: &'a [u8; 15],
+    #[serde_as(as = "Bytes")]
+    #[serde(borrow)]
+    cow_array: Cow<'a, [u8; 15]>,
+    ```
+
+    Note: For borrowed arrays the used Deserializer needs to support Serde's 0-copy deserialization.
+
 ## [1.9.2] - 2021-06-07
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


jonasbb:
Thank you. This definitely is an oversight of mine. For serialization `&Bytes` should work. I'm unsure if that would support 0-copy deserialization, though. The missing implementation on `Cow<'_, [u8; N]>` is a problem :)
It is great that you added the extra documentation.
I can create a new bugfix release later with those changes.

PR #328, opened 2021-06-16
with_prefix macro - Visibility qualifier for generated module

### LINKED ISSUE #327: with_prefix! Making the module public
On a few occasions we have the same struct with the same prefix used in our applications. Currently the `with_prefix` creates a private module for serialize/deserialize and we have to do a hack in order to use the same prefix in other places  and applications.

I'm opening this issue to see how this could be improved in future version.

```rust
// re-export the prefix module!
pub mod prefix_active {
    use serde_with::with_prefix;

    pub use prefix_active::*;

    with_prefix!(prefix_active "active_");
}
```

PS: Thank you all very much for creating and maintaining this crate!

elpiel:
Closes #327 

I'm not sure how to document this and what is the preferred way to add the tests, so I decided to open a PR directly

### DIFF
--- a/src/with_prefix.rs
+++ b/src/with_prefix.rs
@@ -8,7 +8,9 @@ use std::fmt;
 /// Serialize with an added prefix on every field name and deserialize by
 /// trimming away the prefix.
 ///
-/// You can set the visibility of the generated module by prefix the module name with a module visibility.
+/// You can set the visibility of the generated module by prefixing the module name with a module visibility.
+/// `with_prefix!(pub(crate) prefix_foo "foo_");` creates a module with `pub(crate)` visibility.
+/// The visibility is optional and by default `pub(self)`, i.e., private visibility is assumed.
 ///
 /// **Note:** Use of this macro is incompatible with applying the [`deny_unknown_fields`] attribute
 /// on the container.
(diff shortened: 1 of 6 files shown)


src/with_prefix.rs
```diff
@@ -8,6 +8,8 @@ use std::fmt;
 /// Serialize with an added prefix on every field name and deserialize by
 /// trimming away the prefix.
 ///
+/// You can set the visibility of the generated module by prefix the module name with a module visibility.
```
jonasbb, line 11: Maybe you could show an example like `with_prefix!(pub(crate) prefix_foo "foo_");` and mention that it is optional.

jonasbb (changes requested):

jonasbb:
Maybe mention how to use the modifier and that it is optional. Otherwise, the PR looks good. Thanks for creating the PR.

bors delegate=elpiel

src/with_prefix.rs
elpiel: I've used one of the existing prefixes in the example to do so.
Maybe you can be more specific if I have to create a full example like the Match - Player one below or just showcase a non-runnable example here?

jonasbb: Just show a small non-runnable example of the syntax here. Similar to the sentence in the example. Such that you don't have to read the example to know that this is a possibility.

jonasbb:
@elpiel I updated the description. Thanks again for creating the PR. I'm also releasing a new patch version.

elpiel:
That's great, thank you!
Sorry, I didn't have much time to do it, I've been joggling between work and non-work projects 😅

PR #345, opened 2021-08-21
Use collect_str and collect_seq where applicable

jonasbb:
collect_str in particular is nicer than converting to a String first.

bors r+


jonasbb:
Bors didn't start.

bors r+

PR #346, opened 2021-08-22
Use collect_seq and collect_map where applicable v2

dtolnay:
I noticed these comments in #345.

The change in tuple_list_is_map is simply smaller code, but otherwise equivalent.

The change in seq_display_fromstr introduces collect_str to avoid heap allocating a temporary copy of every string object in the sequence.

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -4,7 +4,7 @@ use crate::{utils, Separator};
 use serde::de::{
     Deserialize, DeserializeOwned, Deserializer, Error, MapAccess, SeqAccess, Visitor,
 };
-use serde::ser::{Serialize, SerializeMap, SerializeSeq, Serializer};
+use serde::ser::{Serialize, SerializeSeq, Serializer};
 use std::cmp::Eq;
 #[cfg(doc)]
 use std::collections::{BTreeMap, HashMap};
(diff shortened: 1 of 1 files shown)


jonasbb:
Thank you for the great PR.

`collect_str` seems to be implemented as `self.serialize_str(&value.to_string())` and I only saw it overwritten in `serde_json::Value`, but also with a `to_string`. So I am not sure, if that really saves an allocation.

bors r+

dtolnay:
[Here is where it is overridden for JSON output](https://github.com/serde-rs/json/blob/6346bb30037f31ea69b9a4eb029875e23521681c/src/ser.rs#L445). It does not allocate (on success).

[It is also overridden here where it also saves an allocation](https://github.com/serde-rs/json/blob/d279a78a23819c128bbab99ddc34ef3930ef20b1/src/value/ser.rs#L272). Without this PR serde_with would allocate a string to pass to serialize_str, and Value's serialize_str would allocate another string to put inside the result value. With this PR there is one less string.

jonasbb:
Thanks for pointing out how I'm wrong :) I didn't find the first case via docs.rs and yes, both cases allocate strings, but Value cannot take ownership of the string allocated in serde_with.

PR #360, opened 2021-09-05
Update CI pipeline

jonasbb:
After reading
https://matklad.github.io//2021/09/04/fast-rust-builds.html#ci-workflow
implement some of the CI workflow recommendations.

* Split testing into a `--no-run` step and a testing step, to be able to
  get timing for build and execution.
* Also add `--no-fail-fast` during execution.
* Disable debuginfo during CI. Do this via environment variables instead
  of modifying Cargo.toml

bors merge


jonasbb:
bors merge

PR #362, opened 2021-09-07
unit tests for string_empty_as_none

nklhrstv:
Deserialize `OptionStringEmptyNone` from Some and None too, not just from unit
`visit_some` and `visit_none` implemented for `OptionStringEmptyNone`'s Visitor
unit tests for `string_empty_as_none` added

### DIFF
--- a/tests/rust.rs
+++ b/tests/rust.rs
@@ -6,6 +6,7 @@ use fnv::{FnvHashMap as HashMap, FnvHashSet as HashSet};
 use pretty_assertions::assert_eq;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
+use serde_test::{assert_tokens, Token};
 use serde_with::CommaSeparator;
 use std::cmp;
 use std::collections::{BTreeMap, BTreeSet, LinkedList, VecDeque};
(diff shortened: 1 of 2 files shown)


jonasbb:
Hi @nklhrstv,

thanks for the contribution. I am not sure, that is how the `string_empty_as_none` module should be modified. It is not meant to act like an `Option` for deserialization purposes. It should act more like a `String`.

For customizing deserialization `serde_as` works much better, since here you can compose the desired behavior from multiple parts.
I think you can achieve what you want like this:

```rust
#[serde_as]
#[derive(Deserialize)]
struct Foobar {
	#[serde(default)]
	#[serde_as(deserialize_as = "DefaultOnNull<NoneAsEmptyString>")]
	s: Option<String>,
}
```

The `NoneAsEmptyString` is equivalent to `string_empty_as_none`, the `DefaultOnNull` allows you to deserialize like an `Option`.

Could you please test if that solves your problem? It has the nice added benefit that if you ever want to deserialize a `Vec<Option<String>>` this is easily possible.

nklhrstv:
Hi @jonasbb Thanks for the great example. ive already applied your suggestions and it works exactly how i wanted it to work.
Ive removed everything but the unit tests and renamed to PR.

tests/rust.rs
```diff
@@ -598,6 +598,16 @@ fn tuple_list_as_map_vecdeque() {
     );
 }
 
+#[test]
+fn test_string_empty_as_none() {
+    #[derive(Debug, PartialEq, Deserialize, Serialize)]
+    struct S(#[serde(with = "serde_with::rust::string_empty_as_none", default)] Option<String>);
```
jonasbb, line 604:
```suggestion
    struct S(#[serde(with = "serde_with::rust::string_empty_as_none")] Option<String>);
```

The `default` should not be there, otherwise you tell serde that this field can also be missing.

jonasbb:
Glad to hear that my alternative solution is working for you.

Your PR looks almost good now. Could you please address the one change and squash your commits? The CI failure is unrelated to your changes. I fixed it on master now, so a rebase should fix that.
Thanks for the additional test.

bors delegate

jonasbb:
bors delegate+

PR #367, opened 2021-09-25
Implements support for reference types with skip_serializing_none macro

### LINKED ISSUE #366: skip_serializing_none cannot handle references
Example:

```rust
use serde::Serialize;
use serde_with::skip_serializing_none;

#[skip_serializing_none]
#[derive(Serialize)]
struct Person<'a> {
    pub name: &'a String,
    pub age: &'a Option<u64>,
}

fn main() {
    let name = String::from("test");
    let age = None;
    let t = Person {
        name: &name,
        age: &age,
    };
    println!("{}", serde_json::to_string(&t).unwrap()); // {"name":"test","age":null}
}
```

Changing the `Person` struct to not use references results in the expected behavior.

jmgilman:
Fixes #366.

Adds a bit of pre-work to detect if the field is a reference and, if so, extracts the inner path to use for determining if it's an option. This allows supporting `&Option<...>` as a field type.

### DIFF
--- a/serde_with_macros/src/lib.rs
+++ b/serde_with_macros/src/lib.rs
@@ -297,7 +297,20 @@ pub fn skip_serializing_none(_args: TokenStream, input: TokenStream) -> TokenStr
 
 /// Add the skip_serializing_if annotation to each field of the struct
 fn skip_serializing_none_add_attr_to_field(field: &mut Field) -> Result<(), String> {
-    if let Type::Path(path) = &field.ty {
+    // Extract inner path from references
+    let path = if let Type::Path(p) = &field.ty {
+        Some(p)
+    } else if let Type::Reference(r) = &field.ty {
+        if let Type::Path(p) = &*r.elem {
+            Some(p)
+        } else {
+            None
+        }
+    } else {
+        None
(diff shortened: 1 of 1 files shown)


jonasbb:
Hi, thanks for the idea and the PR. I checked a bit in the GitHub search if this might break some code, since now more fields are skipped, even if they might not be indented to be skipped.
So far, it looks like most cases derive `Deserialize` too, which cannot work on a `&Option`. Other cases have no lifetime parameter, but I think `&'static Option` is unlikely. I found a couple of cases with `Option<&'a T>`, which are unaffected by this change. Overall, I think this has minimal to no impact on existing code.

Could you please add a test for this new functionality and some documentation about this behavior. The tests should go in `serde_with_macros/tests/skip_serializing_null.rs`. The documentation should be updated to include a mention and an example in `serde_with_macros/src/lib.rs` and mention the behavior under `` //! ## `skip_serializing_none` `` in `src/lib.rs` too.

jonasbb:
Due to inactivity, I will close this issue. You are welcome to reopen it if you want to keep working on it.

PR #369, opened 2021-10-01
Base64 Support

### LINKED ISSUE #350: Add support for base64 de/encoding
Support for base64 encoding would look similar to the existing `hex` module. One significant difference is the amount of different base64 character sets. Different character set can be supported using multiple format-like types. Custom character sets would require an approach similar to `StringWithSeparator` of a new trait or full const generics.

jonasbb:
Create a new optional module to house the code.
Add a new `CharacterSet` trait, which is used to encode the different
character sets with multiple unique types.
Add a basic test.

Missing:
* [x] More tests for the other character sets.
* [x] Rustdoc documentation
* [x] Changelog entries
* [x] serde_as Guide entry

Closes #350

### DIFF
--- a/serde_with/CHANGELOG.md
+++ b/serde_with/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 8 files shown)


jonasbb:
bors merge

PR #390, opened 2021-11-06
Remove unmaintained actions

jonasbb:
Most actions in the actions-rs organizations were last updated more than a year ago meaning they are effectively unmaintained.
Remove the actions and call rustup and cargo manually.

Switch to `softprops/action-gh-release` for creating releases, since the current one is no longer maintained.


jonasbb:
bors merge

PR #408, opened 2022-02-02
Ensure that `HashSet` can be used with custom hashers

jonasbb:
https://users.rust-lang.org/t/recursive-serde-deserialize-with-definition/70865/3
A user reported that `FnvHashSet` could not be deserialized in
combination with `serde_as`.
This commit adds the necessary bound on the hasher and adds a test to
prevent regressions.


jonasbb:
bors merge

PR #413, opened 2022-02-09
Allow generics for `SerializeDisplay` and `DeserializeFromStr` derives

jonasbb:
The derives did not take into account that types can have generics. Ensure that lifetime and type generics are added to the generated impls. Add new tests to `serde_with_test` to prevent regressions.


jonasbb:
bors merge

PR #436, opened 2022-04-19
Add tests for IndexMap and IndexSet

jonasbb:
During the tests a missing implementation of `DuplicateInsertsFirstWinsMap` was noticed, but is now included.
`indexmap` was also missing the `serde-1` feature flag.

bors merge


jonasbb:
bors merge

Issue #435, opened 2022-04-19
no_std support


pwnorbitals:
It looks like this crate doesn’t support `no_std`, making it harder to serialize long byte arrays in embedded environments. Is this on the roadmap ?

jonasbb:
Hi @pwnorbitals, `no_std` support is currently not on the roadmap, since nobody requested it before. However, I gladly review PRs adding support and helping implement that feature. I never used `no_std`, so I could really use some help setting up features and testing.

Generally, I don't see a problem with `no_std` support. serde fully supports `no_std` and `rustversion` can be replaced with a build script to check the compiler version. The newly stabilized weak dependency features should also help here. All other crates are already feature gated, so that those implementations should not cause any problems.
Adding a `std` feature and gating existing functionality behind it is as far as I know a breaking change.

---

Could you maybe expand a bit on what exactly you need or in general expect from a `no_std` crate? Do you need `core`, `core+alloc` or both?
Are there any `SerializeAs` implementations more important? I guess, each implementation could be made `no_std` compatible one by one.

jonasbb:
I started working on this in #444. So far I only changed the imports and feature gated imports and implementations. The changes so far seem relatively unproblematic to me and can probably be merged anyway as future preparations.
To really support `no_std` properly are two main questions to answer and some more work updating macros and dependencies.
1. What is the target where to use `no_std`? Is it only `core or `core`+`alloc`.
2. I need help from someone who understand `no_std` how to test it. Otherwise it is not possible to ensure that macros or dependencies do not depend on `std` in some way. I looked at `serde`, but they only have minimal tests for `no_std` and also no dependencies which could interfere.

garbageslam:
I ran into this today too.

Core only is best, Core + alloc will still be fine for many no_std projects. Generally the more you can minimize your dependencies the better it is for your users.

I can try to offer a PR to help with this, I have a lot of experience moving stuff to no_std

garbageslam:
It looks like you already did a lot of work on this in this commit in master, thank you! https://github.com/jonasbb/serde_with/commit/c9d92269fd493b467ae39ee44b1de92c325669a1

At this point you are still bringing in `extern crate std` unconditionally? https://github.com/jonasbb/serde_with/blob/7d9aebc1096903ea4e29ccdb6ae2f95e897cafde/serde_with/src/lib.rs#L267

[...]

PR #444, opened 2022-04-30
Support `no_std` environments with `serde_with`

jonasbb:
This is the initial work to support `no_std` for `serde_with` (#435).

This updates all imports to import from `core` or `alloc` where possible. Additionally, implementations are gated behind the `std` feature.

TODOs:
* [ ] Update macros to import from `core`/`alloc`.
* [ ] Update `serde_with_macros` to generate code referencing `core`/`alloc`.
* [ ] Figure out if `core` or `core`+`alloc` is the better baseline?
* [ ] Update imports to use `no-default-features` and enable `alloc`/`std` as necessary.
* [ ] Figure out how to do testing. This is necessary to ensure the rewritten macros work properly.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -2,5 +2,6 @@
 
 # format_code_in_doc_comments = true
 # format_strings = true
-imports_granularity = "Module"
+imports_granularity = "Crate"
+group_imports = "One"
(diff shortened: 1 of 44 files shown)


jonasbb:
Superseded by #468

PR #446, opened 2022-05-03
Disable the OSSF Scorecard Action

jonasbb:
Without a way to supress analysis results more permanently the security
tab becomes too noisy. Either the action needs to allow an allowlist or
the GitHub UI needs to be better in permanently supressing lines.
The biggest annoyance is that each change to a action tag will trigger
a new warning, even if the same line was ignored before.

https://github.com/ossf/scorecard-action/issues/143


jonasbb:
bors merge

jonasbb:
bors merge

PR #454, opened 2022-05-12
Import from `core` and `alloc` in preparation for #435

jonasbb:
This PR changes the imports to use `core` or `alloc` when possible. It also uses `#![no_std]` but unconditionally imports `std`. This changes the prelude the `core` one, meaning all `alloc` items now must be explicitly imported.

bors merge


jonasbb:
bors merge

Issue #85, opened 2020-02-21
Create a proc-macro attribute which applies a second attribute on all fields of a type.


jonasbb:
Serialization code is often repetitive. One helper to improve on this already exists in this crate, the [`serde_with::skip_serializing_none`](https://docs.rs/serde_with/1.4.0/serde_with/attr.skip_serializing_none.html) attribute.

However, this could be generalized to work with any attribute on a user-specified type (list).
This could look like

```rust
#[apply(Vec, HashMap => #[serde(skip_serializing_if = "$TYPE::is_empty"])]
struct {}
```

This would apply the attribute on all fields of type `Vec` and `HashMap`. On the right hand side, the placeholder `$TYPE` would be replaced by the type of the field. This would make it easier to skip serializing all empty container types.

`serde_with::skip_serializing_none` could in the end be implemented on top of this more general proc-macro.

xmo-odoo:
A supplemental use-case for this is a companion to `skip_serializing_none`: APIs which need that (swagger / openapi using `required` fields) tend to need `skip_serializing_if = "Option::is_none"` on the serialization side, but also `default` on the deserialization side.

`skip_serializing_none` only covers the first one, so when one needs to handle both sides of the serialization  (sometimes on the same struct e.g. because it's defined for both the client and server to use) it can be a bit frustrating.

jonasbb:
> A supplemental use-case for this is a companion to `skip_serializing_none`: APIs which need that (e.g. JSON swagger / openapi) tend to need `skip_serializing_if = "Option::is_none"` on the serialization side, but also `default` on the deserialization side.

You do not need `default` on a field of type `Option`. Missing fields automatically default to `None`. The exception is if you use `#[serde(with = ...)]` or `#[serde_as(...)]`. In the first case using `default` changes the meaning, and the second case is tracked here #185.

> * does serde allow multple `#[serde]` annotations per field? What if a user needs both a blanket annotation, and a specific annotation on a specific field (e.g. `rename`)?

Multiple annotations work flawlessly. Their individual meanings get joined.

> * what about reverting the behaviour for one specific field? e.g. if the user wants to bulk-apply "skip empty collections" but one or two of them must be serialised empty? Or an API which mixes optional fields (skipped) and required nullable fields (not skipped)?

[...]

PR #523, opened 2022-11-03
Add  parametrised version of "skip_serializing_none"

PawelBis:
I have an use case where I need to use `#[serde(skip_serializing_if = "Foo::bar")]` on almost all fields of the struct. Existing `skip_serializing_none` is not good enough.

This mr adds `skip_serializing_if` which is parametrised version of `skip_serializing_none`. In above use case, it would be used as `#[skip_serializing_if("Foo::bar")`. I am not sure if the name I selected is correct (probably not). I was also considering `skip_serializing_when`, but I am up for a discussion.

Additionally we could change `skip_serializing_none` to use `#[skip_serializing_if("Option::is_none")]` under the hood.

### DIFF
--- a/serde_with_macros/src/lib.rs
+++ b/serde_with_macros/src/lib.rs
@@ -221,6 +221,57 @@ where
     }
 }
 
+/// Apply function on every field of structs or enums
+fn apply_function_with_params_to_struct_and_enum_fields<F, T>(
+    input: TokenStream,
+    function: F,
+    params: T,
+) -> Result<TokenStream2, Error>
+where
+    F: Copy,
+    F: Fn(&mut Field, &T) -> Result<(), String>,
+{
+    /// Handle a single struct or a single enum variant
+    fn apply_on_fields<F, T>(fields: &mut Fields, function: F, params: &T) -> Result<(), Error>
+    where
+        F: Fn(&mut Field, &T) -> Result<(), String>,
+    {
+        match fields {
+            // simple, no fields, do nothing
+            Fields::Unit => Ok(()),
+            Fields::Named(ref mut fields) => fields
+                .named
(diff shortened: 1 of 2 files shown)


serde_with_macros/src/lib.rs
```diff
@@ -436,6 +487,243 @@ fn field_has_attribute(field: &Field, namespace: &str, name: &str) -> bool {
...
+/// impl Wrapper {
+///     pub fn is_nothing(&self) -> bool {
+///         match self {
+///             Self::Nothing => true,
+///             _ => false,
+///         }
+///     }
+/// }
+///
+/// #[skip_serializing_if("Wrapper::is_nothing")]
```
jonasbb, line 558: It is no longer necessary for attributes to rely so much on strings. For the most part, any tokens are supported. I think removing the string quotes here would make it nicer. There is also no pre-existing serde interface similar. So confusion or consistency with that does not apply.

jonasbb (commented):
Thank you for providing the PR. This already looks quite good.

I wonder, would you see https://github.com/jonasbb/serde_with/issues/85 as an acceptable alternative to this PR? As I see it, #85 is in large parts more powerful and could be used for the same purposes. This PR shines in being more concise and checking for pre-existing `skip_serializing_if` attributes. Using #85 you can use any expression in the `skip_serializing_if` as it wouldn't need to be bound to the type, since the type is provided separately.

jonasbb:
Just FYI: Using #524 you could write:
```rust
#[serde_with::apply(
	Wrapper => #[serde(skip_serializing_if = "Wrapper::is_nothing")]
)]
```

jonasbb:
This PR is making no progress. You are welcome to reopen the PR if you still need it.

PR #526, opened 2022-11-16
Rename generics in macro outputs to reduce conflicts

Elrendio:
**Good to know:**
- `cargo +nightly fmt` formatted the comments with long lines
- I've made this PR because using `SerializeDisplay` on a struct that was generic over `S` (ex `struct MyStruct<S: Borrow<str>> { ... }` would fail as the struct generic type would conflict with the serialize generic `S`


jonasbb:
Thank you for the PR. I will add a couple of tests to `./serde_with_test/tests/derive_display_fromstr.rs` before publishing a new version. It already contains a case with generics, but none with the used 1-letters here, i.e., `D`, `E`, `S`.

bors r+

PR #536, opened 2022-12-23
Mention "Bytes" on README.md and lib.rs

ryoqun:
I think https://github.com/jonasbb/serde_with/pull/277 can be more promoted. :)

As I wrote the current status of serialization of const generics like bytes (https://github.com/solana-labs/solana/pull/29196#discussion_r1052189935), i think `"Bytes"` is best solution so far.

@jonasbb hey, this functionary saved us. however it took a while to spot this. I think `Bytes` can be promoted. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,6 @@ let arrays: Arrays<100, 128> = Arrays {
     nested: Box::new([[111; 64]; 100]),
     optional: Some([222; 128]),
     bytes: [0x42; 128],
-
 };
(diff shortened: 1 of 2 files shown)


jonasbb:
Hi, I'm glad you found this crate useful :). The lack of const-generic support in `serde_bytes` seems to come up from time to time, so calling out support for it in the README could be useful. Thanks for the PR.

bors r+

PR #543, opened 2023-02-03
Update base64 dependency to 0.21

jeff-hiner:


jonasbb:
Hi, thanks for creating the PR. Sadly, upgrading base64 is a breaking change due to the existence of the [`CharacterSet`](https://docs.rs/serde_with/latest/serde_with/base64/trait.CharacterSet.html) trait. So, I would rather not update the `base64` version for now, see also the PR for the previous version #535. From the `base64` changelog there does not seem to be a strong reason to upgrade.

jeff-hiner:
Would you be open to pulling this in and bumping the semver? I have several projects now that all pull in base64 0.21 via `reqwest` which has already moved on, and I'd like to consolidate versions to cut down on build-time dependencies.

It looks like the public trait's return type does change, but it shouldn't be a breaking change for most people as the macro and import names don't change-- note that I didn't have to touch the tests at all and they still pass. Do you think folks are implementing the pub trait or using it directly? I guess it's available, though not something I'd expect folks to stick their fingers into. Maybe we could hide it or make it opaque?

I also thought about trying to use const generics for clarity, but those are not supported yet for non-integral types (the new `::base64::alphabet::Alphabet` is of course a struct).

jonasbb:
Thanks for the PR again. I know it has been sitting for a while, but I think releasing a new version is probably due. The only change I made is to seal the `Alphabet` trait, such that future updates to base64 are no longer breaking changes. I could not find any evidence of someone using that feature to use custom character sets. So likely nobody is affected anyway.

I probably release v3 over the weekend.

bors r+

PR #548, opened 2023-02-13
Generalize how to implement traits for many collection types.

jonasbb:
Add new macros like `foreach_map` which expand once for each supported
map type, e.g., `BTreeMap`, `HashMap`, `IndexMap`.
They take another macro as input and pass the map as the first argument.

It is used like `foreach_map!(map_impl);` and
`foreach_map!(map_as_tuple_seq);`.

This also helps for future implementations.


jonasbb:
This works well for serializing, where most bounds are simple, but not so well for deserializing. For deserializing, different bounds with different generics are necessary.

bors r+

PR #560, opened 2023-03-03
Support targets without `alloc::sync`/atomic pointers.

### LINKED ISSUE #559: Compilation fails for targets without `alloc::sync`/atomic pointers
## Description

I ran into the issue of getting `unresolved import alloc::sync` when targeting `thumbv6m-none-eabi`.

This is caused by `alloc::sync` being disabled on targets without atomic pointers, as can be seen here:
https://github.com/rust-lang/rust/blob/83b05ef0ee05307143e2d04db401d6085f4759b9/library/alloc/src/lib.rs#L249-L250

vembacher:
## Motivation

This is a fix for the issue described in: #559

## Changes

- I added a check to require `target_has_atomic = "ptr"` for imports and implemenations related to `Arc` and `Weak/WeakArc`.
- That should be sufficient to support such targets without requiring new feature flags.

Closes #559

### DIFF
--- a/serde_with/CHANGELOG.md
+++ b/serde_with/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 * `EnumMap` passes the `human_readable` status of the `Serializer` to more places.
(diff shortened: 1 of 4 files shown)


jonasbb:
Looks good. Thank you for the contribution.

bors r+

PR #571, opened 2023-03-10
Provide length when serializing sequence

### LINKED ISSUE #570: 2.3.0 breaks `HashMap` serialization as `Vec` of pairs with `bincode`
`HashMap` serialization as `Vec` of pairs with `bincode` that worked with `serde_with` 2.2.0 breaks with 2.3.0.

# Reproduction

`Cargo.toml`:
```toml
[package]
name = "playground"
version = "0.1.0"
edition = "2021"

[dependencies]
bincode = "1.3.3"
serde = { version = "1.0.154", features = ["derive"] }
serde_with = "=2.3.0"
```

`src/lib.rs`:
```rs
use serde::Serialize;
use serde_with::serde_as;
use std::collections::HashMap;

#[serde_as]
#[derive(Debug, Default, Serialize)]
struct Foo {
    #[serde_as(as = "Vec<(_, _)>")]
    xs: HashMap<i32, i32>,
}

fn main() {
    bincode::serialize(&Foo::default()).unwrap();
}
```

# Expected output
`cargo run` passes with no output.

# Actual output
`cargo run` fails with:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: SequenceMustHaveLength', src/main.rs:13:41
```

stephaneyfx:
Some serializers (e.g. bincode) require a length to serialize a sequence. This code does what `iterator_len_hint` does (called by `collect_seq` in `serde`).

Resolves #570

### DIFF
--- a/serde_with/tests/serde_as/map_tuple_list.rs
+++ b/serde_with/tests/serde_as/map_tuple_list.rs
@@ -352,3 +352,336 @@ fn test_map_as_tuple_with_nested_complex_type() {
             ]"#]],
     );
 }
+
+#[test]
+fn test_map_as_tuple_list_works_with_serializer_that_needs_length_to_serialize_sequence() {
+    use serde::{
+        ser::{
+            SerializeMap, SerializeSeq, SerializeStruct, SerializeStructVariant, SerializeTuple,
+            SerializeTupleStruct, SerializeTupleVariant,
+        },
+        Serializer,
+    };
+    use std::fmt::{self, Debug, Display};
+
+    #[derive(Debug)]
+    enum TestError {
+        LengthRequired,
+        Other,
+    }
+
+    impl Display for TestError {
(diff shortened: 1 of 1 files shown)


jonasbb:
Hi, thanks for the PR. The correct fix is to revert https://github.com/jonasbb/serde_with/pull/566. I am already on it. I would like to use your test case, since bincode is not a dev-dependency. If you remove the changes to `serde_with/src/ser/impls.rs` I can merge it after the revert.

stephaneyfx:
Sounds good. Thank you. I will merge in your changes and remove my fix to keep only the test case.

stephaneyfx:
Unless I messed up something, this PR should now contain only the test as requested.

jonasbb:
Thank you. This looks good.

bors r+

PR #588, opened 2023-04-05
Require serde 1.0.139

nox:
`serde::de::value::StringDeserializer::new` was introduced in 1.0.139.


jonasbb:
Thanks. This looks good. I can create a patch release for this later.

bors r+

PR #618, opened 2023-07-16
Add FromIntoRef and TryFromIntoRef

oblique:
Conversions from references are rare, however a user who controls the implementation of the type, can implement `Into<T> from &'a MyType` and use this to avoid cloning.

### DIFF
--- a/serde_with/src/de/impls.rs
+++ b/serde_with/src/de/impls.rs
@@ -1719,6 +1719,35 @@ where
     }
 }
 
+impl<'de, T, U> DeserializeAs<'de, T> for FromIntoRef<U>
+where
+    U: Into<T>,
+    U: Deserialize<'de>,
+{
+    fn deserialize_as<D>(deserializer: D) -> Result<T, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        Ok(U::deserialize(deserializer)?.into())
+    }
+}
+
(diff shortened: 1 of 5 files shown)


jonasbb:
Thanks for the contribution. I think I see the utility. Can you please confirm: the new types are almost identical to `FromInto` and `TryFromInto`, except that the `SerializeAs` trait bound does not require `Clone` by instead including the reference in the `Into`/`TryInto` bounds? 
Do you have a use case for which you need this?

However, the “borrow” name part seems wrong here. In the serde ecosystem, borrow usually refers to borrowing during deserialization. The new types do not enable borrowing during serialization either, they only remove the `Clone` requirement. The reason is that `for<'a> &'a T: Into<U>` does not allow `U` to borrow from `T`. What the `for<'a>` means is any lifetime, no matter how short. Therefore, for any lifetime of `U` you can find a smaller `'a`, thus `U` cannot borrow from `T`.
I don't have any name suggestions right now. Do you have any idea?

jonasbb:
Thanks for the info. I like `FromIntoRef`/`TryFromIntoRef` the most. Could you please update the name? The rest looks good. You have tests and documentation written. The only thing missing is a changelog entry, but I can create that when releasing the crate.

oblique:
I changed the names. I also changed the first line of the doc comments. Let me know if you prefer other wording there.

jonasbb:
Thank you for the changes. This looks good now.

bors r+

PR #633, opened 2023-08-03
Properly handle i64::MIN when converting to Duration

### LINKED ISSUE #632: `TimestampSeconds<i64>` panics on `i64::MIN`
**Observed behaviour:**
Attempting to deserialize `i64::MIN` as `TimestampSeconds` panics on UB (underflow).

**Expected behaviour:**
Keep calm and don't panic. Return an error instead.

hollmmax:
fixes #632


jonasbb:
Thank you for resolving the issue. Indeed, the code in this crate shouldn't panic, but for most cases return a serialization error if something happens. The solution with `abs_diff` is nice.

I can release the fix tomorrow.

bors r+

PR #637, opened 2023-08-14
feat: add optional support for hasbrown Map & Set

OliverNChalk:
This PR is not yet ready to merge as I've never setup a `#[no_std]` friendly crate before. Looking for guidance for how to properly setup the Cargo.toml to support this (and perhaps validate this holds true).

### DIFF
--- a/serde_with/Cargo.toml
+++ b/serde_with/Cargo.toml
@@ -54,6 +54,7 @@ chrono = ["chrono_0_4"]
 chrono_0_4 = ["dep:chrono_0_4"]
 guide = ["dep:doc-comment", "macros", "std"]
 hex = ["dep:hex", "alloc"]
+hashbrown_0_14 = ["dep:hashbrown_0_14", "alloc"]
 indexmap = ["indexmap_1"]
 indexmap_1 = ["dep:indexmap_1", "alloc"]
 indexmap_2 = ["dep:indexmap_2", "alloc"]
@@ -66,7 +67,7 @@ time_0_3 = ["dep:time_0_3"]
 base64 = {version = "0.21.0", optional = true, default-features = false}
 chrono_0_4 = {package = "chrono", version = "0.4.20", optional = true, default-features = false, features = ["serde"]}
 doc-comment = {version = "0.3.3", optional = true}
-hashbrown = {version = "0.14.0", optional = true, features = ["serde"]}
+hashbrown_0_14 = {package = "hashbrown", version = "0.14.0", optional = true, default-features = false, features = ["ahash", "serde"]}
 hex = {version = "0.4.3", optional = true, default-features = false}
 indexmap_1 = {package = "indexmap", version = "1.8", optional = true, default-features = false, features = ["serde-1"]}
 indexmap_2 = {package = "indexmap", version = "2.0", optional = true, default-features = false, features = ["serde"]}
@@ -114,9 +115,9 @@ path = "tests/hex.rs"
 required-features = ["hex", "macros"]
 
 [[test]]
-name = "hashbrown"
-path = "tests/hashbrown.rs"
-required-features = ["hashbrown", "macros"]
+name = "hashbrown_0_14"
+path = "tests/hashbrown_0_14.rs"
+required-features = ["hashbrown_0_14", "macros"]
 
 [[test]]
 name = "indexmap_1"
--- a/serde_with/Cargo.toml
+++ b/serde_with/Cargo.toml
@@ -66,6 +66,7 @@ time_0_3 = ["dep:time_0_3"]
 base64 = {version = "0.21.0", optional = true, default-features = false}
 chrono_0_4 = {package = "chrono", version = "0.4.20", optional = true, default-features = false, features = ["serde"]}
 doc-comment = {version = "0.3.3", optional = true}
+hashbrown = {version = "0.14.0", optional = true, features = ["serde"]}
 hex = {version = "0.4.3", optional = true, default-features = false}
 indexmap_1 = {package = "indexmap", version = "1.8", optional = true, default-features = false, features = ["serde-1"]}
 indexmap_2 = {package = "indexmap", version = "2.0", optional = true, default-features = false, features = ["serde"]}
@@ -111,6 +112,11 @@ required-features = ["chrono_0_4", "macros"]
 name = "hex"
 path = "tests/hex.rs"
 required-features = ["hex", "macros"]
+
+[[test]]
+name = "hashbrown"
+path = "tests/hashbrown.rs"
+required-features = ["hashbrown", "macros"]
 
 [[test]]
 name = "indexmap_1"
--- a/serde_with/Cargo.toml
+++ b/serde_with/Cargo.toml
@@ -67,7 +67,7 @@ time_0_3 = ["dep:time_0_3"]
 base64 = {version = "0.21.0", optional = true, default-features = false}
 chrono_0_4 = {package = "chrono", version = "0.4.20", optional = true, default-features = false, features = ["serde"]}
 doc-comment = {version = "0.3.3", optional = true}
-hashbrown_0_14 = {package = "hashbrown", version = "0.14.0", optional = true, default-features = false, features = ["ahash", "serde"]}
+hashbrown_0_14 = {package = "hashbrown", version = "0.14.0", optional = true, default-features = false, features = ["serde"]}
 hex = {version = "0.4.3", optional = true, default-features = false}
 indexmap_1 = {package = "indexmap", version = "1.8", optional = true, default-features = false, features = ["serde-1"]}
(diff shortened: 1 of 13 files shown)


serde_with/src/de/impls.rs
```diff
@@ -33,6 +37,11 @@ macro_rules! foreach_map_create {
             HashMap<K: Eq + Hash, V, S: BuildHasher + Default>,
             (|size| HashMap::with_capacity_and_hasher(size, Default::default()))
         );
+        #[cfg(feature = "hashbrown")]
+        $m!(
+            HashbrownMap<K: Eq + Hash, V, S: BuildHasher + Default>,
+            (|size| HashbrownMap::with_capacity_and_hasher(size, Default::default()))
+        );
```
OliverNChalk, line 44: Flagging that I followed the `indexmap` template but I have not properly setup the dependencies to enable hashbrown in a `no_std` environment (assuming that's even possible). Rather than take a punt I thought I would wait for your guidance on the matter @jonasbb

serde_with/Cargo.toml
```diff
@@ -65,6 +65,7 @@ time_0_3 = ["dep:time_0_3"]
 base64 = {version = "0.21.0", optional = true, default-features = false}
 chrono_0_4 = {package = "chrono", version = "0.4.20", optional = true, default-features = false, features = ["serde"]}
 doc-comment = {version = "0.3.3", optional = true}
+hashbrown = {version = "0.14.0", optional = true, features = ["serde"]}
```
jonasbb, line 68: The crate name should be versioned. If in the future hashbrown v0.15 comes out, it should be possible to support it, without extra troubles. By adding the version into the imported name it is much simpler to support multiple versions of the same dependency.

None of the default features seem important, so I would disable them all. All implementations in serde_with should be independent of the hasher if possible, thus no need for `ahash`. `inline-more` is only a performance optimization and can be enabled by downstream crates. `allocator-api2` seems only relevant for raw types.

```suggestion
hashbrown_0_14 = {version = "0.14.0", optional = true, default-features = false, features = ["serde"]}
```

You should add a matching feature above too. Hashbrown seems to require the `alloc` crate, so alloc support should be enabled whenever using hashbrown too.

```toml
hashbrown_0_14 = ["dep:hashbrown_0_14", "alloc"]
``````

File: serde_with/src/guide/feature_flags.md [hunk 3]
```diff
@@ -6,17 +6,19 @@ Each entry will explain the feature in more detail.
...
-3. [`base64`](#base64)
-4. [`chrono_0_4`](#chrono_0_4)
-5. [`guide`](#guide)
-6. [`hex`](#hex)
-7. [`indexmap_1`](#indexmap_1)
-8. [`indexmap_2`](#indexmap_2)
-9. [`json`](#json)
-10. [`macros` (default)](#macros-default)
-11. [`time_0_3`](#time_0_3)
+- [Available Feature Flags](#available-feature-flags)
```
jonasbb, line 9: Can you please leave the features as a numbered list.

serde_with/src/de/impls.rs
```diff
@@ -1,4 +1,6 @@
 use crate::{formats::*, prelude::*};
+#[cfg(feature = "hashbrown")]
+use hashbrown::{HashMap as HashbrownMap, HashSet as HashbrownSet};
```
jonasbb, line 3: Maybe it is better to add the version number into the renames here too, similar to the `indexmap_2` case. You would need to propagate that to the `de` and `ser` modules.

```suggestion
use hashbrown::{HashMap as HashbrownMap014, HashSet as HashbrownSet014};
```

serde_with/src/de/impls.rs
jonasbb: I see nothing that would prevent the `no_std` usecase. hashbrown does not even have a `std` feature, probably because it is so deeply embedded in the Rust standard library.

You can change this line in the CI tests to include the `hashbrown_0_14` feature. That will compile the code on a target without std support.
https://github.com/jonasbb/serde_with/blob/fc30b63125c501b12e1c9e53e50a7618c4180b47/.github/workflows/ci.yaml#L92C157-L92C157

serde_with/Cargo.toml
```diff
@@ -111,6 +112,11 @@ name = "hex"
 path = "tests/hex.rs"
 required-features = ["hex", "macros"]
 
+[[test]]
+name = "hashbrown"
```
jonasbb, line 116: A version number for the test here would be great too.

jonasbb (commented):
Thanks for the PR. Overall the implementations look good. Please use the version number of the dependency in all the places. That makes it much simpler to add support for hashbrown v0.15 and any version afterwards. This means the crate and feature should be `hashbrown_0_14`, the type import renames and tests ideally too.
I left some more detailed remarks

serde_with/src/guide/feature_flags.md
OliverNChalk: Damn, I manually renumbered this, I must have some mark down formatter running on save.

serde_with/Cargo.toml
OliverNChalk: If we drop the `ahash` dependency, I believe we must declare our own default hasher. Some questions I have that you may already know the answer to:

- Should we use ahash as default.
- Should we use the default hasher from std::collections::HashMap.
  - If we choose this one, is that hasher available in #[no_std]?
- Does our selection of default hasher affect downstream users who would expect ahash as the default, I would assume not and that the macros would run on their namespace and thus use their hashers?

jonasbb: There should be no need to declare a default hasher. Rather, the code should be generic to work with most/any kind of hasher and should not assume anything about it. At least for serialization, this works flawlessly. Iterating over a HashMap does not require hashing anything, so no bounds need to be placed. For deserializing hashing is needed. But the HashMap only really cares that it can get a `Hasher`, which is what the `BuildHasher` trait gives. This leaves the question of how to create the `BuildHasher`. The simplest solution is `Default`. That will not always work, but most implement `Default`. That is why the deserialization code has `S: BuildHasher + Default` as the bounds.

This might leave the tests without a defined hasher. But all tests only run on targets with std so the `std::collections::hash_map::DefaultHasher` type can be used there.

Any user of hashbrown and serde_with can still specify the hasher they want and that can be ahash.

OliverNChalk: Looking at `hashbrown` it appears disabling `ahash` leaves `HashMap` and `HashSet` without a default hash builder:
https://github.com/rust-lang/hashbrown/blob/b2190e22e82b0b3fbae14f508bf5fc08a977df80/src/map.rs#L17

I also am unaware of how we can work around this short of a PR to `hashbrown` allowing `RandomState` to be used as a default hasher (but that would break the additive nature of feature flags so unlikely they would accept it).

As such, it stands to reason we cannot compile the hashbrown_0_14 without the `ahash` flag, is my assessment correct?

jonasbb: No code in serde_with should make any assumption about what the default hash builder is. All implementations should be generic over that parameter. As such, removing that should not cause any problems and if it does, then the implementations need to be fixed.

The one exception to that are the tests. Here you do need concrete instances of hashmaps, which means you need a concrete hasher too. The tests only run on targets with std available. The dev-dependencies contain [`fnv`](https://docs.rs/fnv) too. Between `std::collections::hash_map::DefaultHasher` and `fnv::FnvHasher` there should be no need for a third hasher. If the implementations work with both hashers, the implementations are sufficiently generic to support ahash too.

Maybe I am missing something here. Buy my plan would be to remove the feature. See if the implementations in serde_with still compile. In the tests it might be necessary to specify either of the above hasher manually.

OliverNChalk: If you remove the feature, serde_with no longer compiles. My belief is its because the fallback default hasher is bogus:
https://github.com/rust-lang/hashbrown/blob/b2190e22e82b0b3fbae14f508bf5fc08a977df80/src/map.rs#L17

Perhaps I've missdiagnosed, are you able to checkout and build without ahash?

jonasbb:
Ok, I found some issues with the `foreach_*!` macros in the `de` module. They did not work with `hashbrown` without `ahash`. The used macros failed to forward the extra type parameters and constrains to the implementations, meaning the implementations were only available with the default hasher type for each map and set. I updated the macros to make it work without the `ahash` dependency. The tests are using `fnv` instead of `ahash`, which changes the serialization order in some cases.

You can review the changes again if you like, mainly b20796807db6bba7a6bded6daa4bc01a61eb58e9. I think this is now fine to merge. You can merge it, or I will to it, after you had some time to check it. Thank you for the contribution. This really helped with fixing the macros, since any wrong assumption about the hasher argument lead to a compilation failure, instead of silently working like with the other maps.

bors delegate+

jonasbb (approved):

OliverNChalk:
I will let you handle the merge but happy with the changes and having learnt something about macros

Issue #651, opened 2023-10-06
Question about minimum serde version


banool:
Hey hey!

I see that currently `serde_with` requires a minimum serde version of `1.0.157`. I wonder if any earlier version would work or if there is some feature you need from that version onward?

I ask because I have a codebase which is unfortunately stuck on `1.0.149` and I want to use a crate that depends on `serde_with`.

Thanks!

jonasbb:
Version 1.0.49 is already 5 years old. I am not sure any version of serde_with ever supported that. serde does add new API surfaces occasionally, and that is used. Supporting 1.0.49 is not possible. The last bump was to force serde to use syn v2. But this PR https://github.com/jonasbb/serde_with/pull/588 bumps the dependency to fix the minimal version specified, and that requires 1.0.139. So my best guess is anything under 1.0.139 cannot be supported right now, but it might be higher.

banool:
Oh my sorry, I meant 1.0.149, my mistake.

banool:
I can test this btw, this week I'll try pinning the version to 1.0.149 to see if it builds.

banool:
PR to decrease the version: https://github.com/jonasbb/serde_with/pull/653/. Everything seems to pass.

jonasbb:
PR is merged

banool:
Could you release a new version of the crate so this change gets picked up?

PR #653, opened 2023-10-09
Decrease minimum serde version to 1.0.152

banool:
As far as I can tell, the minimum version of 1.0.157 is unnecessary. The way I've determined that is just from running the build + tests:
```
cargo build --workspace
cargo test --workspace
```

This addresses https://github.com/jonasbb/serde_with/issues/651. For my needs at least 1.0.152 is actually sufficient (rather than 1.0.149), though we could try go lower.


jonasbb:
Loosening the version constraint looks fine, as long as the test still pass. Thanks for the PR.
bors r+

PR #655, opened 2023-10-12
Add support for core::ops::Bound

qsantos:
I have encountered a case where I would like to use `Bound<Cow<'a, str>>` and borrow the string from the deserialized buffer instead of copying. To use `BorrowStr`, I had to add support for `core::ops::Bound` to `serde_with`.

### DIFF
--- a/serde_with/src/ser/impls.rs
+++ b/serde_with/src/ser/impls.rs
@@ -130,6 +130,24 @@ where
     }
 }
 
+impl<T, U> SerializeAs<Bound<T>> for Bound<U>
+where
+    U: SerializeAs<T>,
+    T: Sized,
+{
+    fn serialize_as<S>(source: &Bound<T>, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match source {
+            Bound::Unbounded => Bound::Unbounded,
+            Bound::Included(ref v) => Bound::Included(SerializeAsWrap::<T, U>::new(v)),
+            Bound::Excluded(ref v) => Bound::Excluded(SerializeAsWrap::<T, U>::new(v)),
+        }
+        .serialize(serializer)
+    }
+}
+
 #[cfg(feature = "alloc")]
 impl<T, U> SerializeAs<Rc<T>> for Rc<U>
 where
--- a/serde_with/tests/serde_as/lib.rs
+++ b/serde_with/tests/serde_as/lib.rs
@@ -29,6 +29,7 @@ use alloc::{
     sync::{Arc, Weak as ArcWeak},
 };
 use core::cell::{Cell, RefCell};
+use core::ops::Bound;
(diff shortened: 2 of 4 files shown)


serde_with/src/ser/impls.rs
```diff
@@ -130,6 +130,31 @@ where
...
+    where
+        S: Serializer,
+    {
+        match source {
+            Bound::Unbounded => Bound::Unbounded,
+            Bound::Included(ref v) => Bound::Included(SerializeAsWrap::<T, U>::new(v)),
+            Bound::Excluded(ref v) => Bound::Excluded(SerializeAsWrap::<T, U>::new(v)),
+        }
+        .serialize(serializer)
+        /*
```
jonasbb, line 148: Can you please remove the commented code

serde_with/tests/serde_as/lib.rs
```diff
@@ -136,6 +137,41 @@ fn test_option() {
     );
 }
 
+#[test]
+fn test_bound() {
+    #[serde_as]
+    #[derive(Debug, Serialize, Deserialize, PartialEq)]
+    struct S(#[serde_as(as = "_")] Bound<u32>);
```
jonasbb, line 144: This does not test your code, but instead the already existing Serialize and Deserialize implementations for Bound. You want `as = "Bound<_>"`.

Can you please add a test that converts the inner type somehow? For example `Bound<DisplayFromStr>`.

jonasbb (changes requested):

jonasbb:
Thanks for the PR. It makes sense to support Bounds. The code needs some updates, especially the tests. The trait implementations look OK.

serde_with/tests/serde_as/lib.rs
qsantos: > This does not test your code, but instead the already existing Serialize and Deserialize implementations for Bound. You want `as = "Bound<_>"`.

Oh, that makes sense! Done.

By the way, wouldn't it be the same with [`test_option`](https://github.com/jonasbb/serde_with/pull/655/files#diff-db1bb1fa4de5910d07d9a9d12ae3276fda201487e23f3006933052fac23613adR119) then?

> Can you please add a test that converts the inner type somehow? For example `Bound<DisplayFromStr>`.

Done.

serde_with/src/ser/impls.rs
qsantos: Sorry about that! Done.

qsantos:
Thanks for taking the time to review my changes. I have applied your suggestions.

serde_with/tests/serde_as/lib.rs
jonasbb: Yes, you are right. That test looks wrong too.

jonasbb:
Thanks for the changes. It looks good now.
bors r+

PR #666, opened 2023-12-07
Include support for schemars

swlynch99:
I saw the existing PR for this over in #623 and figure that since I'm interested in using `schemars` and `serde_with` together then I might as well push this over the finish line.

What's been kept the same as #623:
- It introduces a top level type `Schema<T, TA>`
- It implements `JsonSchema` for `Schema<T, TA>`
- The `#[serde_as]` macro can add `#[schemars(with = "Schema<T, ...>")]` attributes to fields annotated with `#[serde_as(as = ...)]`

What's new here:
- A whole bunch more `JsonSchema` impls for various std wrapper types.
- Impls of `JsonSchema` for `Schema<T, Same>` and `Schema<T, DisplayFromStr>`
- `#[serde_as]` will now automatically detect a `#[derive(JsonSchema)]` attribute and use that to determine if it should add `#[schemars(with = ...)]` to fields. This only happens if the `schemars_0_8` feature is enabled and can be overridden by doing `#[serde_as(schemars = true/false)]`.

I have a branch where I'm working to implement `JsonSchema` for all the types in this crate but figured it would be best to make multiple (still large) PRs instead of one enormous one.


jonasbb:
Thank you for picking up that PR :) Overall, this already looks good. Some more tests would be good. I will need some more time to look at the PR, but I can't give you a timeline right now.

If I see it correctly, the code is even supposed to handle conditional implementations of `JsonSchema` with the `SchemaFieldConfig::Conditional` variant. That is cool. This will need a test too.

Do you have any suggestion how best to test the new schema implementations? There are two basic tests types I can think of: 1) Generate a schema for a type and verify that it is as expected (like the current test) and 2) generate a schema and serialize a value to JSON, then validate the JSON using the schema.

swlynch99:
No rush on the PR review :) I figured that this might take a while.

I have added tests for the handling of conditional `JsonSchema` derives as well as for validating the schema for `[T; N]` which is the only implementation that doesn't just forward to an existing one. I have also moved some of my helper functions over from [my other branch with all the impls](https://github.com/swlynch99/serde_with/tree/schemars-full).

As for testing, I see three different kinds of tests that can be done. You've already got the first two
1. These are basically snapshot tests. They are the easiest to write but I think that they are also the least useful. IMO the only real case where these are useful is if you find reviewing the change schema more helpful than reviewing the code. If you do think it's worth it to have a bunch of these then I would suggest using a snapshot testing library to do so (e.g. [insta](https://insta.rs/)).
2. Checking the serialized json against the schema is good as a sanity check. It'll check the happy path but, if the schema is implemented right, then it should never fail so it ends up being somewhat limited.
3. The third one is to validate custom json against the schema. This is the only way to validate that the schema excludes everything it is supposed to.

I would lean towards a mix of 2 and 3 when writing tests. For more complicated impls I think it should mostly be type 3 tests.

I do think, though, that it is reasonable to skip the test cases when the `JsonSchema` implementation is just forwarding to another and that mapping is obvious. What should be considered "obvious" is a judgement call but I think pretty much all of the impls for `std`'s types fit. I also think forwarding rules like `Same` -> `T` or `DisplayFromStr` -> `String` are fairly obvious as well.

Ultimately, though, the decision on what should have tests is up to you :)

Now it looks like I need to figure out the right dependency version to add to get a `Cargo.lock` that is compatible with 1.64.0.

jonasbb:
> As for testing, I see three different kinds of tests that can be done. You've already got the first two
> 
>     1. These are basically snapshot tests. They are the easiest to write but I think that they are also the least useful. IMO the only real case where these are useful is if you find reviewing the change schema more helpful than reviewing the code. If you do think it's worth it to have a bunch of these then I would suggest using a snapshot testing library to do so (e.g. [insta](https://insta.rs/)).
> 
>     2. Checking the serialized json against the schema is good as a sanity check. It'll check the happy path but, if the schema is implemented right, then it should never fail so it ends up being somewhat limited.
> 
>     3. The third one is to validate custom json against the schema. This is the only way to validate that the schema excludes everything it is supposed to.
> 
> 
> I would lean towards a mix of 2 and 3 when writing tests. For more complicated impls I think it should mostly be type 3 tests.

I agree with this assessment. I think type 1 and 2 work for more basic implementations that basically just forward the definition to another type. You can quickly check that `Option<(Vec<A>, [(B, C, D]; 5])>` produces a schema that works. The snapshots you can spot check that they look plausible. Type 2 works then for more complex variants where the generated schema is too large, but then requires multiple type instances for testing.

Your type 3 is good. I forgot that there are cases that do not delegate to another schema. For these more complex implementations, it makes sense to have positive and negative cases.

So far, I have used [`expect_test`](https://docs.rs/expect-test/latest/expect_test/) for snapshot testing, since I didn't anything advanced yet. The snapshots are quite stable overall.

> I do think that it is reasonable to skip the test cases when the `JsonSchema` implementation is just forwarding to another and that mapping is obvious. What should be considered "obvious" is a judgement call but I think pretty much all of the impls for `std`'s types fit. I also think forwarding rules like `Same` -> `T` or `DisplayFromStr` -> `String` are fairly obvious as well.

I would not totally skip tests for them, but they can be reduced. For example, you could have one test with `Option<(Vec<A>, [(B, C, D]; 5])>`, that way you test `Option`, tuple, `Vec`, and arrays at once.

---

If the current MSRV is making trouble for you, feel free to bump it.

swlynch99:
I've gone and added snapshot tests that cover most of the forwarding impls for std types. These are done through a macro so it should be pretty easy to write more snapshot tests in the future.

---
The MSRV is no issue. Generating the lockfile was inconvenient since I had to use the latest rust nightly for the msrv-aware resolver but once it is generated then everything works as normal.

PR #679, opened 2024-01-19
Allow external crates to add their own `JsonSchema` impls for custom `SerializeAs` and `DeserializeAs` types

swlynch99:
This is written on top of #676.

---

Implementing `JsonSchema` directly on `serde_with::Schema<T, TA>` works fine when within `serde_with`. Once you are outside the crate, however, the orphan rules prevent you from adding impls of the form
```rust
impl JsonSchema for Schema<T, MyCustomType> { ... }
```
We would like users to be able to make that impl (or something equivalent) so we need to find a way to make it work. We make this work by adding a new layer of trait indirection.

This PR introduces a new `JsonSchemaAs<T>` trait, adds a blanket
```rust
impl<T, TA> JsonSchema for Schema<T, TA> where TA: JsonSchemaAs<T>
```
and then changes all the existing impls to be for `JsonSchemaAs<T>` instead of on `Schema<T, TA>`.

Now, when a user wants to add json schema support for their own type they can implement `JsonSchemaAs<T>` for their custom type and the orphan rules will allow it.

As a bonus, this seems to have made the extraordinarily confusing ["reached recursion limit"](https://github.com/rust-lang/rust/issues/119839) errors that I have been encountering disappear. Which is nice.

### DIFF
--- a/serde_with/src/lib.rs
+++ b/serde_with/src/lib.rs
@@ -331,7 +331,7 @@ pub mod json;
 mod key_value_map;
 pub mod rust;
 #[cfg(feature = "schemars_0_8")]
-mod schemars_0_8;
+pub mod schemars_0_8;
 pub mod ser;
 #[cfg(feature = "std")]
 mod serde_conv;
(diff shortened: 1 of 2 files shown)


swlynch99:
This should be a bit easier to review now that I've rebased onto the latest master :)

PR #683, opened 2024-01-22
Add `JsonSchema` impls for `BytesOrString` and `BoolFromInt`

swlynch99:
This PR continues where #679 left off. Since the impls get more complicated as we go on I will be chunking them up into smaller PRs so that reviewing them stays manageable. I also don't plan on doing more than 1 or maybe 2 PRs concurrently. I just considered #682 important enough that it should jump the queue. 

## Details
- `JsonSchemaFor` impls for both `BytesOrString` and `BoolFromInt`
- Tests for both of the above

## Notes
- `BytesOrString`'s schema is represented as an enum that can be either a string or an array of bytes. The one difference is that the string representation is marked as "writeOnly" as it is never emitted.
- I originally also did something similar for `BoolFromInt<Flexible>` but I didn't think that this was useful compared to the extra confusion when reading the schema. I have changed it so it's schema is just that of an integer.

### DIFF
--- a/serde_with/src/schemars_0_8.rs
+++ b/serde_with/src/schemars_0_8.rs
@@ -401,24 +401,9 @@ impl JsonSchemaAs<bool> for BoolFromInt<Flexible> {
         "serde_with::BoolFromInt<Flexible>".into()
     }
 
-    fn json_schema(gen: &mut SchemaGenerator) -> Schema {
+    fn json_schema(_: &mut SchemaGenerator) -> Schema {
         SchemaObject {
             instance_type: Some(InstanceType::Integer.into()),
-            subschemas: Some(Box::new(SubschemaValidation {
-                any_of: Some(std::vec![
-                    gen.subschema_for::<WrapSchema<bool, BoolFromInt<Strict>>>(),
-                    SchemaObject {
-                        instance_type: Some(InstanceType::Integer.into()),
-                        metadata: Some(Box::new(Metadata {
-                            write_only: true,
-                            ..Default::default()
-                        })),
-                        ..Default::default()
-                    }
-                    .into()
-                ]),
-                ..Default::default()
-            })),
             ..Default::default()
         }
         .into()
--- a/serde_with/src/schemars_0_8.rs
+++ b/serde_with/src/schemars_0_8.rs
@@ -6,12 +6,15 @@
 //! see [`JsonSchemaAs`].
 
 use crate::{
-    formats::Separator,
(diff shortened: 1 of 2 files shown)


serde_with/src/schemars_0_8.rs
```diff
@@ -362,6 +365,55 @@ impl<T> JsonSchemaAs<T> for DisplayFromStr {
...
+    }
+
+    fn schema_id() -> Cow<'static, str> {
+        "serde_with::BoolFromInt<Strict>".into()
+    }
+
+    fn json_schema(_: &mut SchemaGenerator) -> Schema {
+        SchemaObject {
+            instance_type: Some(InstanceType::Integer.into()),
+            number: Some(Box::new(NumberValidation {
```
jonasbb, line 380: I think this would be better modelled as an enum with the two values 0 and 1. A number with min and max would allow the value 0.5, I think.

serde_with/src/schemars_0_8.rs
```diff
@@ -374,6 +426,42 @@ impl<T> JsonSchemaAs<T> for Bytes {
...
+
+    fn json_schema(gen: &mut SchemaGenerator) -> Schema {
+        SchemaObject {
+            subschemas: Some(Box::new(SubschemaValidation {
+                any_of: Some(std::vec![
+                    gen.subschema_for::<Vec<u8>>(),
+                    SchemaObject {
+                        instance_type: Some(InstanceType::String.into()),
+                        metadata: Some(Box::new(Metadata {
+                            write_only: true,
```
jonasbb, line 446: You use the `write_only` to indicate that deserializing will accept a `String`, but serialization will never produce it, right?

jonasbb (changes requested):

serde_with/src/schemars_0_8.rs
swlynch99: Normally you would be right, but the `instanceType: "integer"` part means that 0.5 would not be valid.

This is a good test case to have in case it gets modified in the future though. I have gone and added some tests to that effect.

serde_with/src/schemars_0_8.rs
swlynch99: Yup! That is correct.

serde_with/src/schemars_0_8.rs
jonasbb: Yes, you are right :) 0.5 is not an integer

jonasbb (approved):

PR #694, opened 2024-02-08
Eliminate dependency on serde's "derive" feature

dtolnay:
This allows serde, and other crates downstream of it like serde_json, to compile in parallel with serde_derive. Without this PR, in any dependency graph that pulls in serde_with, first serde_derive would need to compile, then serde would need to compile, then serde_json would need to compile.

https://github.com/sharkdp/bat/pull/2815 has a diagram of the impact of this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -861,6 +861,7 @@ dependencies = [
  "schemars",
  "serde",
  "serde-xml-rs",
+ "serde_derive",
  "serde_json",
  "serde_test",
  "serde_with_macros",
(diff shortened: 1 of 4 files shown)


dtolnay:
CI failures look unrelated to this change. See #695.

jonasbb:
Thank you for the PRs. This looks ok, but I think it re-introduces the risk of mismatching serde dependency versions. Bumping the serde dependency to v1.0.186 should mitigate this. I can bump the dependency and release a new patch version.

dtolnay:
Thank you, that would be great. Yes, serde 1.0.186+ contains https://github.com/serde-rs/serde/pull/2588 which eliminates the version mismatch possibility.

jonasbb:
v3.6.1 is published now

PR #697, opened 2024-02-10
Implement `JsonSchemaAs` for `EnumMap`

swlynch99:
It's been a little busy but I'm back with yet another PR :)

This is the first one that really involves a custom deserialize implementation. `EnumMap` involves transforming a list of enums to a map and to generate its schema we end up needing to do the same to the schema.

Basically, we end up transforming
```json
{
  "type": "object",
  "oneOf": [
    { "properties": { "A": "..." } },
    { "properties": { "B": "..." } }
  ]
}       
```
into
```json
{
  "type": "object",
  "properties": {
    "A": "...",
    "B": "..."
  }
}
```

I have also included a number of tests that I think should cover most of the different cases.

## Notes
- If the schema for the inner type doesn't match what would be expected for a rust enum then we'll end up generating an invalid overall schema. I don't think this is an issue since it doesn't appear that `EnumMap` itself works in this case?

### DIFF
--- a/serde_with/src/schemars_0_8.rs
+++ b/serde_with/src/schemars_0_8.rs
@@ -518,6 +518,58 @@ where
     forward_schema!(WrapSchema<BTreeMap<K, V>, BTreeMap<KA, VA>>);
 }
 
+impl<T> JsonSchemaAs<Vec<T>> for EnumMap
+where
+    T: JsonSchema,
+{
+    fn schema_name() -> String {
+        std::format!("EnumMap<{}>", T::schema_name())
+    }
+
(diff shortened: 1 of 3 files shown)


jonasbb:
Hey, thank you for that. The transformation looks good. I think it is fine to not handle schema mismatches for the inner enum. At best, you could detect that something is going on and raise an error. The `json_schema` function is infallible, so it is not possible to nicely pass an error. Panicing would be a worse choice. `EnumMap` can only handle default style enums. I don't think there is much error handling in that case either, since nothing can be done really.

PR #713, opened 2024-03-07
Implement JsonSchemaAs for KeyValueMap

swlynch99:
I'm back with yet another PR! This time we're doing `KeyValueMap`.

I think this one is probably the most complicated conversion that needs to be done to make the schema work. We need to convert a schema that looks like this
```json
{
  "type": "object",
  "properties": {
    "$key$": "...",
    "a": "...",
  },
  "required": ["$key$", "a"]
}
```
into this
```json
{
  "type": "object",
  "additionalProperties": {
    "type": "object",
    "required": ["a"],
    "properties": { "a": "..." }
  }
}
```

On its own that is not to bad but it gets more complicated since we also have to deal subschemas which can be generated from types that make use of `#[serde(untagged)]` and `#[serde(flatten)]`. See the doc comment on `kvmap_transform_schema` for more details.

# Notes
- I'm not entirely sure this is correct in the presence of an [`anyOf`](https://json-schema.org/draft/2020-12/json-schema-core#section-10.2.1.2) subschema that makes use of `minProperties` and `maxProperties` (normal ones where they have a `$key$` field should work fine). I'm not entirely sure there is one correct behaviour in the case either but to me this seems like a edge case that is pretty far out there so it should be fine.
- It also isn't correct if the `$key$` property ends up being defined in some external schema (i.e. `$ref` points to some random document on the internet). This seems like an edge case and I don't think there's any reasonable behaviour here anyway.

### DIFF
--- a/serde_with/src/schemars_0_8.rs
+++ b/serde_with/src/schemars_0_8.rs
@@ -12,10 +12,14 @@ use crate::{
 use ::schemars_0_8::{
     gen::SchemaGenerator,
     schema::{
-        ArrayValidation, InstanceType, Metadata, NumberValidation, Schema, SchemaObject,
-        SubschemaValidation,
+        ArrayValidation, InstanceType, Metadata, NumberValidation, ObjectValidation, Schema,
+        SchemaObject, SingleOrVec, SubschemaValidation,
     },
     JsonSchema,
+};
+use core::{
+    mem::ManuallyDrop,
+    ops::{Deref, DerefMut},
 };
 
 //===================================================================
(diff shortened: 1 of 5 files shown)


swlynch99:
The CI errors all seem to be related to the `#[warn(unused_qualifications)]` lint in `Cargo.toml`. That doesn't seem related to this PR and when I went to fix them the suggestions from `cargo clippy --fix` seemed like they would be invalid under `#[no_std]`. I'll leave this on up to you to fix and/or disable the lints and then rebase once that is done.

serde_with/src/schemars_0_8.rs
```diff
@@ -895,3 +1076,52 @@ forward_duration_schema!(TimestampSecondsWithFrac);
...
+impl<T, F: FnOnce(T)> DerefMut for DropGuard<T, F> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.value
+    }
+}
+
+impl<T, F: FnOnce(T)> Drop for DropGuard<T, F> {
+    fn drop(&mut self) {
+        // SAFETY: value is known to be initialized since we only ever remove it here.
+        let value = unsafe { ManuallyDrop::take(&mut self.value) };
```
jonasbb, line 1121: Small nit: I think you can use a `Option` for the value. Then you `take()` the option in the `drop` implementation. The `Deref` implementation would need a match with panic, in case the `Option` is `None`, but it should be possible to get rid of the `None`.

jonasbb:
Thank you, this looks good. Your comments are good, explain what is going on, and I could follow them. Thanks for getting this to work :)

serde_with/src/schemars_0_8.rs
swlynch99: I'm occasionally a little prone to using `unsafe` for stuff like this. If you would prefer that I use an option here I can definitely submit a PR to change that.

PR #715, opened 2024-03-07
Avoid emitting `with` when there's already a `schema_with` annotation

### LINKED ISSUE #712: Unable to use `serde_with::serde_as` in conjunction with `schemars(schema_with)`
I am trying to properly encode `Base64` encoded data using schemars, but I am failing while the `schemars_0_8` feature is enabled. `Base64` does not implement `JsonSchemaAs`, so I went ahead and tried to implement it myself using the `schema_with` annotation.

```rust
#[serde_with::serde_as]
#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
#[serde(rename_all = "camelCase")]
pub struct Image {
    #[serde_as(as = "serde_with::base64::Base64")]
    #[cfg_attr(feature = "schemars", schemars(schema_with = "base_64_schema"))]
    pub buffer: Vec<u8>,
}

#[cfg(feature = "schemars")]
fn base_64_schema(_: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
    // These properties are valid JSON-schema properties, but they are not supported by schemars
    let mut extensions = std::collections::BTreeMap::new();
    extensions.insert(
        "contentEncoding".to_owned(),
        serde_json::Value::from("base64"),
    );
    extensions.insert(
        "contentMediaType".to_owned(),
        serde_json::Value::from("image/*"),
    );

    schemars::schema::SchemaObject {
        instance_type: Some(schemars::schema::InstanceType::String.into()),
        extensions,
        ..schemars::schema::SchemaObject::default()
    }
    .into()
}
```

This fails with the error: schemars attribute cannot contain both `with` and `schema_with`

any help would be immensely appreciated. Thank you so much for the great library and overall integration of `schemars` into it!

swlynch99:
I spotted #712 and since I had finally gotten around to making another set of PRs I figured I might as well submit a fix.

This treats a `#[schemars(schema_with = "...")]` attribute the same as a `#[schemars(with = "...")]` attribute when deciding whether to emit our own.

There is also a new test case that validates that everything works as expected.

Fixes #712

### DIFF
--- a/serde_with/CHANGELOG.md
+++ b/serde_with/CHANGELOG.md
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 * Implement `JsonSchemaAs` for `EnumMap` by @swlynch99 (#697)
 
+### Fixed
+
+* Detect conflicting `schema_with` attributes on fields with `schemars` annotations by @swlynch99 (#715)
+    This extends the existing avoidance mechanism to a new variant fixing #712.
+
 ## [3.6.1] - 2024-02-08
 
 ### Changed
--- a/serde_with/tests/schemars_0_8.rs
+++ b/serde_with/tests/schemars_0_8.rs
@@ -117,6 +117,41 @@ fn schemars_custom_with() {
     }));
 }
 
+#[test]
+fn schemars_custom_schema_with() {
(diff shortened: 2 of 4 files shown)


jonasbb:
Thank you for fixing this issue. I saw the existing test for the other attributes. You planned already ahead for everything here :)

PR #719, opened 2024-03-11
Implement `JsonSchemaAs` for `OneOrMany`

swlynch99:
Here's the next type. This one implements `JsonSchemaAs` for `OneOrMany`. It is fortunately much simpler than `KeyValueMap`.

For `OneOrMany` the transformation we want to do is basically to take this schema
```json
{ "$ref": "..." }
```

and transform it into
```json
{
  "anyOf": [
    { "$ref": "..." },
    {
      "type": "array",
      "items": { "$ref": "..." }
    }
  ]
}
```

that is, we change it so it can take either an array or a single value. The only difference between the `PreferOne` and the `PreferMany` variants is that the inner object is marked as `writeOnly: true`.

I have also included some tests that should cover pretty much everything.

### DIFF
--- a/serde_with/src/schemars_0_8.rs
+++ b/serde_with/src/schemars_0_8.rs
@@ -6,7 +6,7 @@
 //! see [`JsonSchemaAs`].
 
 use crate::{
-    formats::{Flexible, Format, Separator, Strict},
+    formats::{Flexible, Format, PreferMany, PreferOne, Separator, Strict},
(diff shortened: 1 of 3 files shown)


jonasbb:
That looks great :) Thank you.

jonasbb:
I published your new `JsonSchemaAs` implementations and fix for the `serde_as` macro in v3.7.0

PR #721, opened 2024-03-28
Implement `JsonSchemaAs` for `PickFirst`

swlynch99:
This is the second-to-last type that I've got an implementation queued up for. The only one left after this is `Seq`.

This one is pretty straightforward: we put the involved schemas in an `anyOf` subschema. All the schemas except the first are annotated with `writeOnly: true` to indicate that they are only accepted when deserializing and not emitted when serializing. The one extra detail is that I've used an internal `allOf` in each element so that setting `writeOnly: true` works even if the schema of the type is a reference to an external schema.

I have also included tests and a snapshot test to show that everything works.


jonasbb:
Thanks, the implementation looks good :) Like all your PRs

PR #733, opened 2024-04-20
Implement (De)Serialization for Pinned Smart Pointers

Astralchroma:
Implements `SerializeAs` and `DeserializeAs` for `Pin<Box<T>>`, `Pin<Rc<T>>`, and `Pin<Arc<T>>`

### DIFF
--- a/serde_with/CHANGELOG.md
+++ b/serde_with/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+* Implement (De)Serialization for Pinned Smart Pointers by @Astralchroma (#733)
+
 ## [3.7.0] - 2024-03-11
 
 ### Added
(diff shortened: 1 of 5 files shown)


Astralchroma:
Seems the lints are bugged and are flagging the macros as unused when they are in fact used. I'll go ahead and add an `#[allow(unused-macros)]` for now.

jonasbb:
Hi, thank you for submitting the PR. Everything looks good here. I extended the `SerializeAs` side with support for `Pin<&'a T>` and `Pin<&'a mut T>`.

I wonder if it is possible to support any `Pin<T>`, not just with specific pointer types. `Pin::as_ref` almost looks like it could work, but then is also return `Deref::Target` and I couldn't figure out how to make that work. On the deserialization side we need the well-known pointer types, since only they have safe conversions to `Pin`.

I leave this PR open for a day or two, in case you have an idea how to extend the serialization side, otherwise this looks good to merge.

Astralchroma:
Yea, honestly not sure about how to go about extending it to work using Deref stuff. Had a go at it myself and couldn't really get very far with it. Might come back to it later as it seems like an interesting idea, however I'm just going to leave it for now I think.

PR #760, opened 2024-06-28
Implement JsonSchemaAs for OneOrMany instead of JsonSchema

swlynch99:
Currently, there is an `impl JsonSchema for WrapSchema<OneOrMany<...>>`. This works fine if you only directly need `OneOrMany` but causes errors if you want to nest it (e.g. `Option<OneOrMany<_>>`).

This PR changes the impl to a `JsonSchemaAs` impl, which is the correct approach. I think I just missed these ones during a rebase. In any case, the change should be fully backwards compatible due to the blanket schema impl.

I have added a snapshot test. The actual snapshot doesn't matter - the real test is that the struct below compiles

```rust
#[serde_as]
#[derive(Serialize, Deserialize, JsonSchema)]
struct Test {
    #[serde_as(as = "Option<OneOrMany<_>>")]
    optional_vec: Option<Vec<String>>,
}
```


jonasbb:
Thanks, that looks good :)

PR #763, opened 2024-07-03
Implement `MapSkipError`, analogous to `VecSkipError`, but for map-like data structures.

johnmave126:
Currently, there is a `VecSkipError` for resiliently handling heterogenous sequences. However, there are also heterogenous maps in the wild, notably the bundle version information plist from Apple.

This PR implements `MapSkipError`, analogous to `VecSkipError`, deserializes an entry only if both the key and value are deserializable.

## Notable Changes
- The impl of `VecSkipError` is moved to a new file `de/skip_error.rs` and `ser/skip_error.rs` to reuse `GoodOrError` with `MapSkipError`.
- The impl of `MapSkipError` works for all the maps enumerated by `impls::macros::foreach_map`.

## Unanswered Questions
- [x] Do we also need to implement this for kv-tuple lists?
- [x] Do we also need to implement modules under `serde_with::rust` for `BTreeMap` and `HashMap`?


serde_with/src/guide/serde_as_transformations.md
```diff
@@ -181,6 +182,25 @@ colors: Vec<Color>,
 // => vec![Blue, Green]
 ```
 
+## De/Serialize into a map, ignoring errors
+
+[`MapSkipError`]
+
+For formats with heterogeneously typed maps, we can collect only the elements where both key and value are deserializable.
+This is also useful in conjunction to `#[serde(flatten)]` to ingore some entries when capturing additional fields.
```
jonasbb, line 190:
```suggestion
This is also useful in conjunction to `#[serde(flatten)]` to ignore some entries when capturing additional fields.
```

jonasbb (approved):

jonasbb:
Thanks for the PR, this looks all really good.

Regarding your questions:
1. It is not necessary to support kv-list. You can add support if you want, but this PR is fine to merge as is.
2. No, you shouldn't add more to `serde_with::rust`. The `VecSkipError` doesn't have any other variations either. In general, it is difficult to combine such different behaviors and if you combine too many you have a giant compatibility matrix.

johnmave126:
Cooool.

You are right. Under the current way of enumerating different types of map, adding support to kv-list probably means adding support for `MapPreventDuplicates` and alike, which is out of the scope of this PR. I don't plan to further change this PR then.

jonasbb:
Thanks for the work :) I try to prepare a new release this weekend.

PR #777, opened 2024-09-05
Add newline separators

jayvdb:


jayvdb:
ping @jonasbb

jonasbb:
> Unfortunately this doesnt work:
> 
> ```rust
>     #[serde_as]
>     #[derive(Debug, Deserialize, PartialEq, Serialize)]
>     #[serde(untagged)]
>     enum NewlineSeparatedStringSet {
>         Unix(#[serde_as(as = "StringWithSeparator::<UnixLineSeparator, String>")] BTreeSet<String>),
>         Dos(#[serde_as(as = "StringWithSeparator::<DosLineSeparator, String>")] BTreeSet<String>),
>     }
> ```

serde always processes untagged enums top down. Since the unix line ending is part of the dos line ending, this leads to the wrong matches that you reported. Switching the order, i.e., having dos first, should work. Assuming the individual strings do not contain any `\r` or `\n`.

> This is sufficient for my needs as I only want Unix style multi-line strings parsed, but it would be nice if a cross-platform "new line" separator could be created. I suspect this would need a new interface.

Do you mean something that is always the native newline separator of the current platform? That should be doable. If you want something that can match any separator while splitting strings (i.e., automatically figuring out which separator is used) then this will not work with the current interface.

jonasbb:
This looks nice, thank you. Newline separators seem a quite stable separator.

PR #778, opened 2024-09-06
Fix typos

jayvdb:


.typos.toml
```diff
@@ -0,0 +1,2 @@
+[default.extend-words]
+Algorythm = "Algorythm"
```
jayvdb, line 2: with this https://crates.io/crates/typos-cli passes without error

happy to add it to the CI if you like

jonasbb: I'm not sure about the configuration yet, but what is the difference to this?

```
[default.extend-identifiers]
TheAlgorythm = "TheAlgorythm"
```

jonasbb:
This looks good. Thank you for the PR. The `typos` tool looks interesting. Automating it makes sense too, but I wonder if adding it as a pre-commit hook would work better.

PR #782, opened 2024-09-21
feat(serde_with_macros): proper handling of `cfg_attr` and `schemars`

sivizius:
With this commit, `cfg_attr` will be handled properly. The current implementation does not accept e.g.:

```rust
#[serde_as]
#[derive(JsonSchema, Serialize)]
struct Test {
    #[serde_as(as = "DisplayFromStr")]
    #[cfg_attr(any(), arbitrary("some" |weird| syntax::<bool, 2>()))]
    #[cfg_attr(any(), schemars(rename = "foo"), schemars(range(min = 0)))]
    custom: i32,
}
```

Because due to https://github.com/jonasbb/serde_with/blob/master/serde_with_macros/src/utils.rs#L91, the parser does not accept `cfg_attr` with multiple arguments, which is allowed, see e.g. https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute. Another issue I’ve found with `arbitrary` where some unexpected errors due to this custom parser: https://github.com/jonasbb/serde_with/blob/master/serde_with_macros/src/utils.rs#L191-L218 in combination with this line: https://github.com/jonasbb/serde_with/blob/master/serde_with_macros/src/utils.rs#L240.

Line 240 will only `continue`, if the outer-most meta if neither `cfg_attr` or `schemars`. If however the attribute is `cfg_attr`, the code after the `match`-block is reached and the parser will just assume, that the inner meta is `schemars` and thus fail to parse any meta with some weird syntax.

Even `#[schemars(length(min = 1.0))]` will fail, because the parser expects all `schemars`-attributes to be a list of key-value-pairs (`foo = bar`), thus failing for `length(min = 1.0)`, even though this is perfectly legal (https://graham.cool/schemars/deriving/attributes/#length).

I introduced the type `LazyBool` to simplify dealing with various conditions.

### DIFF
--- a/serde_with/tests/schemars_0_8.rs
+++ b/serde_with/tests/schemars_0_8.rs
@@ -95,6 +95,22 @@ fn schemars_basic() {
 }
 
 #[test]
+fn schemars_other_cfg_attrs() {
+    #[serde_as]
+    #[derive(JsonSchema, Serialize)]
+    struct Test {
+        #[serde_as(as = "DisplayFromStr")]
+        #[cfg_attr(any(), arbitrary("some" |weird| syntax::<bool, 2>()))]
+        #[cfg_attr(any(), schemars(with = "i32"))]
+        custom: i32,
+    }
+
+    check_matches_schema::<Test>(&json!({
+        "custom": "23",
+    }));
+}
+
+#[test]
 fn schemars_custom_with() {
     #[serde_as]
     #[derive(JsonSchema, Serialize)]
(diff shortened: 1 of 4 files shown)


jonasbb:
Thanks for the PR. In my opinion, it is a bit unfortunate that it is necessary to parse the `cfg_attr` in the first place, but without that you loose some usability. Since the PR is not so big and this fixes some issues I am fine to merge this. I didn't really check the code in detail, but it looks reasonable overall. Let's hope the test cover enough.

PR #793, opened 2024-10-10
feat(serde_with): make `BytesOrString` adjustable

sivizius:
At the moment, any `Vec<u8>` annotated with `#[serde(as = "BytesOrString")]` will always be serialised as an array of integers in the range `0–255`. However, often these bytes are valid strings, the internal representation as `Vec<u8>` is often just an implementation detail. This PR adds a generic type parameter `PREFERENCE`, which must implement the new trait `TypePreference`, which is a subtrait of `SerializeAs<[u8]>`. This parameter defaults to the new marker-strut `PreferBytes`, which serialises `Vec<u8>` as an array like before. Two additional marker-structs are `PreferString`, which tries to convert `&[u8]` to `&str` first and will fallback to serialising as array only if the bytes do not represent a valid string, as well as `PreferAsciiString`, which additionally checks that all characters of the string are valid ASCII-characters, otherwise falling back to the array as well.

### DIFF
--- a/serde_with/src/formats.rs
+++ b/serde_with/src/formats.rs
@@ -80,6 +80,59 @@ create_format!(
     Unpadded
 );
 
+/// When serializing a value of a type,
+///   that allows multiple types during deserialization,
+/// prefer a specific type.
+pub trait TypePreference: SerializeAs<[u8]> {}
+
+/// Prefer serializing it as ASCII string.
+pub struct PreferAsciiString;
+
+impl TypePreference for PreferAsciiString {}
+
+impl SerializeAs<[u8]> for PreferAsciiString {
+    fn serialize_as<S>(source: &[u8], serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match core::str::from_utf8(source) {
+            Ok(text) if text.is_ascii() => serializer.serialize_str(text),
+            _ => serializer.serialize_bytes(source),
+        }
+    }
+}
+
+/// Prefer serializing it as bytes.
+pub struct PreferBytes;
+
+impl TypePreference for PreferBytes {}
+
+impl SerializeAs<[u8]> for PreferBytes {
+    fn serialize_as<S>(source: &[u8], serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_bytes(source)
+    }
+}
+
+/// Prefer serializing it as string.
+pub struct PreferString;
+
+impl TypePreference for PreferString {}
+
+impl SerializeAs<[u8]> for PreferString {
+    fn serialize_as<S>(source: &[u8], serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match core::str::from_utf8(source) {
+            Ok(text) => serializer.serialize_str(text),
+            _ => serializer.serialize_bytes(source),
+        }
+    }
+}
+
 /// Specify how lenient the deserialization process should be
 ///
 /// Formats which make use of this trait should specify how it affects the deserialization behavior.
--- a/serde_with/tests/serde_as/lib.rs
+++ b/serde_with/tests/serde_as/lib.rs
@@ -30,7 +30,7 @@ use core::{
 use expect_test::expect;
 use serde::{Deserialize, Serialize};
 use serde_with::{
-    formats::{CommaSeparator, Flexible, Strict},
+    formats::{CommaSeparator, Flexible, PreferAsciiString, PreferString, Strict},
     serde_as, BoolFromInt, BytesOrString, DisplayFromStr, IfIsHumanReadable, Map,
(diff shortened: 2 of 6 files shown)


serde_with/tests/serde_as/lib.rs
```diff
@@ -477,7 +477,7 @@ fn test_none_as_empty_string() {
 }
 
 #[test]
-fn test_bytes_or_string() {
+fn test_bytes_or_string_as_bytes() {
     #[serde_as]
     #[derive(Debug, Serialize, Deserialize, PartialEq)]
     struct S(#[serde_as(as = "BytesOrString")] Vec<u8>);
```
jonasbb, line 483: Can you please add a test for an explicit `PreferBytes` too?

serde_with/src/formats.rs
```diff
@@ -80,6 +80,59 @@ create_format!(
     Unpadded
 );
 
+/// When serializing a value of a type,
+///   that allows multiple types during deserialization,
+/// prefer a specific type.
+pub trait TypePreference: SerializeAs<[u8]> {}
```
jonasbb, line 86: That's neat to just reuse the `SerializeAs` trait for the behavior.

Does the trait need to be fully public? We can't make it private, since it appears in public bounds, but seal all implementations. Similar to https://github.com/jonasbb/serde_with/blob/bc2063440da5168963285196e0345e82ab9ccc4e/serde_with/src/base64.rs#L160-L168.

I would like to keep the serialization and deserialization behavior matching if possible. For that a fully public trait feels unnecessary. Having ASCII and unicode strings supported feels sufficient for now. If it turns out that more freedom is necessary, it can be unsealed later.

jonasbb (changes requested):
Thanks for the PR. This looks really good already :) The addition looks reasonable to me.

jonasbb:
Given this PR is #794 really necessary? The `Bytes` type does not really have any string association, but is primarily to enable calling the (potentially) optimized `serialize_bytes` function.

sivizius:
> Given this PR is #794 really necessary? The `Bytes` type does not really have any string association, but is primarily to enable calling the (potentially) optimized `serialize_bytes` function.

It was more like an obvious addition to this PR. However, making this trait work for `Bytes` as well somewhat changed the serialisation behaviour of `BytesOrString` for the `PreferBytes` case: It will call `serialize_bytes`. Is this a breaking change?

jonasbb:
> It will call `serialize_bytes`. Is this a breaking change?

Oh yeah, that is changing the behavior. In JSON it is not visible, as there is no efficient way to store binary data. There are some tests for `Bytes` using `assert_tokens`.

sivizius:
> > It will call `serialize_bytes`. Is this a breaking change?
> 
> Oh yeah, that is changing the behavior. In JSON it is not visible, as there is no efficient way to store binary data. There are some tests for `Bytes` using `assert_tokens`.

It’s not breaking for `Bytes`, I adjusted the `SerdeAs<[u8]>` implementation of `PreferBytes` to match the current behaviour. It has however changed it for `BytesOrString` though, which used `source.serialize(serializer)` before and uses `serializer.serialize_bytes(source)` now.

serde_with/src/formats.rs
```diff
@@ -80,6 +80,59 @@ create_format!(
...
+impl TypePreference for PreferAsciiString {}
+
+impl SerializeAs<[u8]> for PreferAsciiString {
+    fn serialize_as<S>(source: &[u8], serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match core::str::from_utf8(source) {
+            Ok(text) if text.is_ascii() => serializer.serialize_str(text),
+            _ => serializer.serialize_bytes(source),
```
jonasbb, line 100: As discussed, these `serialize_bytes` should just be `serialize`.

serde_with/src/formats.rs
```diff
@@ -80,6 +80,59 @@ create_format!(
...
+pub struct PreferBytes;
+
+impl TypePreference for PreferBytes {}
+
+impl SerializeAs<[u8]> for PreferBytes {
+    fn serialize_as<S>(source: &[u8], serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_bytes(source)
```
jonasbb, line 115:
```suggestion
        serializer.serialize(source)
```

serde_with/src/formats.rs
```diff
@@ -80,6 +80,59 @@ create_format!(
...
+impl TypePreference for PreferString {}
+
+impl SerializeAs<[u8]> for PreferString {
+    fn serialize_as<S>(source: &[u8], serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match core::str::from_utf8(source) {
+            Ok(text) => serializer.serialize_str(text),
+            _ => serializer.serialize_bytes(source),
```
jonasbb, line 131:
```suggestion
            _ => serializer.serialize(source),
```

PR #797, opened 2024-11-01
feat(serde_with): add `with_suffix!`

### LINKED ISSUE #381: Add `with_suffix` macro?
Perhaps not such a common use case as stripping prefixes, but it would be nice if there was a symmetrical `with_suffix` macro that could strip suffixes.

fgardt:
This should properly close #381.

Apart from the used example it is pretty much identical to the existing `prefix_with!` code, apart for the few places that actually do the suffixing / trimming.


jonasbb:
Thank you for doing the work of adding `with_suffix!`. Since it is mostly a copy, I don't see a problem here. I like that you added a proper new example to the documentation :)

fgardt:
any ETA for a new release that would contain this? @jonasbb

Issue #798, opened 2024-11-04
Generated schemas are not valid when using Rust style names in ref


waltronix:
Not all generated schemas are valid. Especially when using the Rust style names with `<`, `>`, or `, ` for schema name generation it breaks schema validation.

I suggest to add a validation check to `declare_snapshot_test` and then change the generated names #799

waltronix:
The problem does not come from the `,` itself, but from the space after it.

jonasbb:
Closed with #799

PR #799, opened 2024-11-04
Generated schemas are not valid when using rust style names in ref

waltronix:
#798

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,7 +60,7 @@ jobs:
         # It is good to test more than the MSRV and stable since sometimes
         # breakage occurs in intermediate versions.
         # IMPORTANT: Synchronize the MSRV with the Cargo.toml values.
-        rust: ["1.67", "1.70", "1.75", "1.80", "stable", "beta", "nightly"]
+        rust: ["1.70", "1.75", "1.80", "stable", "beta", "nightly"]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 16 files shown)


swlynch99:
I'm not a maintainer of serde-with, just the person who originally wrote the schemars integration. So feel free to ignore this if you want.

With that said, it would be really helpful to document this on the `schema_id` and `schema_name` methods of the `JsonSchemaAs` trait. That the characters need to be valid as a URI path segment for refs to work is very much not apparent unless you are familiar with the json schema spec.

jonasbb:
Hi, thanks for the PR. The change looks fine overall. Adding the extra validation check to the test makes sense.
Do you have a reference which names are allowed? In #798 you listed the three characters `<`, `>`, and `,`. The `,` is still used here and seems to pass the tests.

If the names are invalid anyway, then backwards compatability is not a problem.

swlynch99:
I think it needs to be a URI path segment, so it would be defined in [RFC 3986 section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3). That refers back to section 2 and [section 2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) seems to indicate that all 3 of `(`, `,`, and `)` are valid.

waltronix:
Thanks for your fast response!

I had a mistake in my bug report. There is no problem with the `,` itself, the problem arises from the space after it.

I also extended the doc comment for `schema_name()`. As `schema_id()` is not part of the generated schema according to current documentation I don't think have to require it to be a URI path segment.

PR #803, opened 2024-11-28
Reformat all cargo.toml

nyurik:
Just a minor cleanup to keep spacing consistent


nyurik:
seems your CI really doesn't like me :(

jonasbb:
Thank you for the PR. It is useful. :) Yeah, the CI was broken by the Rust 1.83 update and the coverage testing, but that is unrelated to the changes here.

PR #832, opened 2025-04-29
makes unwrap_or_skip compatible with references (zero-copy)

beroal:
`rust::unwrap_or_skip` was incompatible with references (zero-copy). See examples of this incompatibility in the new tests.


jonasbb:
Thanks for the contribution :) That looks like a sensible change, since the extra bound isn't necessary. I just fixed the formatting of the code and added a changelog entry.

PR #833, opened 2025-05-01
Add `SerializeDisplayAlt` macro

Bryley:
This PR introduces the `SerializeDisplayAlt` derive macro.

This is an alternative to the `SerializeDisplay` macro except instead of using the plain formatting like `format!("{}", ...)`, it serializes with the `Formatter::alternate` flag set to true, like `format!("{:#}", ...)`.


jonasbb:
Thanks for the contribution :) Since it is a pretty small change I don't mind merging this.

PR #847, opened 2025-06-05
Display the underlying error into base64 try_into failures

dlowe:
Using `#[serde_as(as = "Base64")]` on a type which has non-trivial validation in the `TryFrom<Vec<u8>>` implementation, I found that `Base64` is currently "eating" the underlying errors.

This (`Display` the underlying error into the new error) is the simplest fix that works for me. I'm not attached to the specifics, as long as there's some way for me to get at the underlying error, so lmk if you'd prefer a different approach!

### DIFF
--- a/serde_with/src/base64.rs
+++ b/serde_with/src/base64.rs
@@ -113,6 +113,7 @@ const PAD_INDIFFERENT: ::base64::engine::GeneralPurposeConfig =
 impl<'de, T, ALPHABET, FORMAT> DeserializeAs<'de, T> for Base64<ALPHABET, FORMAT>
 where
     T: TryFrom<Vec<u8>>,
+    T::Error: Display,
     ALPHABET: Alphabet,
     FORMAT: formats::Format,
 {
@@ -125,6 +126,7 @@ where
         impl<T, ALPHABET> Visitor<'_> for Helper<T, ALPHABET>
         where
             T: TryFrom<Vec<u8>>,
+            T::Error: Display,
             ALPHABET: Alphabet,
         {
             type Value = T;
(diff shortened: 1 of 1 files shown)


jonasbb:
I understand that debugging might be hampered by swallowing the error message. The problem though is that your code adds a new trait bound `T::Error: Display` to the implementation. If there would be some way to get some useful behavior without any new trait bounds, I will happily merge that. But that sounds a lot like specialization, which doesn't exist yet in Rust.

dlowe:
> I understand that debugging might be hampered by swallowing the error message. The problem though is that your code adds a new trait bound `T::Error: Display` to the implementation.

Yep! On the other hand, this is a _very_ common trait bound for errors—among other things, it's a constraint on the `std::error::Error` trait (so it's ubiquitous in `std`... and also the basis of `thiserror` and the `serde` error traits!) I think it's quite reasonable to add, here.

> If there would be some way to get some useful behavior without any new trait bounds, I will happily merge that. But that sounds a lot like specialization, which doesn't exist yet in Rust.

Without *any* trait bounds, I believe it's impossible. Serde deserializing imposes `serde::de::Error`, which doesn't offer anything more general than [the `Display` constraint on custom()](https://docs.rs/serde/latest/serde/de/trait.Error.html#tymethod.custom).

dlowe:
> While it is quite common, it is not universal. And there is even one counterexample in std that is relevant for here:
> [`impl<T, const N: usize> TryFrom<Vec<T>> for Box<[T; N]>`](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html#impl-TryFrom%3CVec%3CT%3E%3E-for-Box%3C%5BT;+N%5D%3E) uses as its error type `Vec<T>`, where `Display` is never provided, not even conditionally.

Fair enough. Two other ideas:
1. the `Display` constraint could conditional (globally, at compile-time) on a crate feature.
2. a separate struct `Base64WithErrorPropagation` (or whatever) that can live side-by-side with the unconstrained `Base64` in the crate, with docs explaining why there are two.

jonasbb:
> the `Display` constraint could conditional (globally, at compile-time) on a crate feature.

Features need to be additive, meaning that enabling a feature will not break other peoples code. This does not work with adding a new constraint.

I prefer not having a basically identical `Base64WithErrorPropagation` struct next to the existing `Base64`. You are welcome to copy the `Base64` code and add it to your own crates.

PR #849, opened 2025-06-05
Add support for schemars v0.9

### LINKED ISSUE #842: Support for schemars 0.9.0 / 1.0.0
schemars 0.9.0 [has been released](https://github.com/GREsau/schemars/releases/tag/v0.9.0) which is a precursor to 1.0.0.

**serde_with** currently only has support for 0.8.0. When 0.9.0 is supported, minimal changes can be expected once 1.0.0 is released.

swlynch99:
This is mostly a copy-paste from the support for schemars v0.8. There are a couple of changes which make the v0.9 support a bit more messy, but otherwise it's not bad. This commit also duplicates the existing tests, which all pass. Once schemars v1.0 gets released it should hopefully require fewer changes to add support again.

The big changes in schemars that cascade out to support here are
- Schema is now just a very thin wrapper around serde_json::Value. This is rather convenient for defining schemas, less convenient for the transformations we need to make here.
- Some of the trait methods on the JsonSchema trait have had their definitions change.

It's possible the schemars::transform module could be used to make this cleaner. I haven't looked too deeply into what that would involve.

Fixes #842

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,7 +60,7 @@ jobs:
         # It is good to test more than the MSRV and stable since sometimes
         # breakage occurs in intermediate versions.
         # IMPORTANT: Synchronize the MSRV with the Cargo.toml values.
-        rust: ["1.71", "1.75", "1.80", "1.85", "stable", "beta", "nightly"]
+        rust: ["1.74", "1.75", "1.80", "1.85", "stable", "beta", "nightly"]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -942,7 +942,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f55c82c700538496bdc329bb4918a81f87cc8888811bd123cf325a0f2f8d309"
 dependencies = [
  "dyn-clone",
- "schemars_derive",
+ "schemars_derive 0.8.17",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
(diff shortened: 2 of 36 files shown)


swlynch99:
CI seems to be failing since schemars v0.9 has a MSRV of 1.74. What's the best way to make this work? I'm guessing this is the first dependency that has a no versions compatible with 1.71

jonasbb:
Thank you for preparing this :) Feel free to bump the MSRV to 1.74. Here is an example of such a previous change: https://github.com/jonasbb/serde_with/commit/2c6c51bf53212ea2153dc47f3d933533a73754e4

swlynch99:
I've bumped MSRV to 1.74 and added appropriate release notes.

jonasbb:
Thank you for the great PR. This looks mostly like mechanical changes to adapt to the new `JsonSchema` and removal of `SchemaObject` by replacing it with the `serde_json::Value` type.
* Including `String` -> `Cow<'static, str>` transition
* `is_referenceable` -> `inline_schema` transition by negating the default value
* Lots of `json_schema!` instead of `SchemaObject`
* Feature gating some implementations behind `std` since schemars v0.9 only requires `alloc`.

Two notes:
* https://github.com/jonasbb/serde_with/blob/a6806b2874512b78cfd8c56d17cab760362a4add/serde_with/tests/schemars_0_8.rs
    This test file seems to have been skipped. I try converting it for this PR.
* [`c14ced8` (#849)](https://github.com/jonasbb/serde_with/pull/849/commits/c14ced8811a92993956039076badbfda8011bbf8)
    This seems like a pre-existing issue with the schemars v0.8 implementation. I guess that should be fixed for that too.

PR #866, opened 2025-07-07
Allow for schema derivation for the Hex type

0xOmarA:
## Description

This PR allows for the `serde_with::hex::Hex` type to be used with `schemars`

## Tests

Updated the `schemars_basic` test cases with tests for lowercase and uppercase hex schemas


jonasbb:
Thank you for the contribution. This all looks good as far as I can tell. Give that both lowercase and uppercase use the same pattern, does it make sense to merge the two implementations? I am thinking of just making the format generic and not including it in the schema name.

jonasbb:
Thank you, this looks great now :)

PR #878, opened 2025-09-02
feat(serde_with): VecSkipError and MapSkipError error inspection

michelhe:
Added an optional type parameter for error inspection, the use-case is to be able to trace all failed elements and alert the users of the bad input.

I added doctests which should cover the feature well, let me know if I should also have a dedicated unittest for this.


jonasbb:
Thank you for this great PR. It has all the necessary things: clean code, updated docs, even with examples :) It is a relatively minor change, but adding the ability to extract some information is neat.

I only updated the PR to fix a conflict in the changelog and extended the changelog entry a bit.

PR #939, opened 2026-04-28
Add `skip_serializing_default` macro helper

howardjohn:
This helps like the Option one but for other types like Vec, wrappers around Option, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ Foo {a: None, b: None, c: None, d: Some(4), e: None, f: None, g: Some(7)}
 
 If many fields should be omitted whenever they equal their default value, putting the annotations on each field can become tedious.
 The `#[skip_serializing_default]` attribute must be placed *before* the `#[derive]`.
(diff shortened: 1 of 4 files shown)


jonasbb:
Hi, thank you for the contribution. This was previously suggested #410.

Two points from there that I think are worth considering:
1. `Default::default() == Default::default()` is not guaranteed to be true. There is a reason that `std` doesn't have a `is_default` function. `std` even has types where it does not hold.
2. It might be worth to consider `#[serde(default = "some_func")]` annotations as part of this, such that serializing and deserializing are symmetric in the behavior.

jonasbb:
I'm going to close this for now. For me, the question is still, if this can even be reasonably implemented at all, given that `Default::default() == Default::default()` is not guaranteed to be true. There is a reason that std doesn't have a `is_default` function. `std` even has types where it does not hold.

In the meantime, you can try using [`serde_with::apply`](https://docs.rs/serde_with/latest/serde_with/attr.apply.html).

```rust
#[serde_with::apply(
    _ => #[serde(skip_serializing_if = "is_default")],
)]
```

PR #943, opened 2026-05-02
feat: base58

mitinarseny:
This PR adds `base58` feature to encode/decode bytes as base58 strings:
```rust
use serde_with::base58::{Base58, Flickr};

#[serde_as]
struct B58 {
    // The default is the same as Standard character se
    #[serde_as(as = "Base58")]
    default: Vec<u8>,
    // Change the character set
    #[serde_as(as = "Base58<Flickr>")]
    charset_flickr: Vec<u8>,
}

let b58 = B58 {
    default: b"Hello World".to_vec(),
    charset_flickr: b"Hello World".to_vec(),
};
let json = serde_json::json!({
    "default": "JxF12TrwUP45BMd",
    "charset_flickr": "iXf12sRWto45bmC",
});

// Test serialization and deserialization
assert_eq!(json, serde_json::to_value(&b58).unwrap());
assert_eq!(b58, serde_json::from_value(json).unwrap());
```

Please note that there is no support for [checksum](https://en.bitcoin.it/wiki/Base58Check_encoding) functionality, since it complicates the code much. However, it can be added in the future if needed in a backwards-compatible way: just add another type parameter for `Base58` adapter along with its default type.


mitinarseny:
Hey @jonasbb, thanks for this awesome crate!
Do you think this PR has a chance to be merged? 🥹

serde_with/Cargo.toml
```diff
@@ -159,6 +163,7 @@ time_0_3 = ["dep:time_0_3"]
 
 # When adding new optional dependencies, update the documentation in feature-flags.md
 [dependencies]
+bs58 = { version = "0.5.1", optional = true, default-features = false }
```
jonasbb: This should be sorted alphabetically

serde_with/Cargo.toml
```diff
@@ -37,11 +37,11 @@ default = ["std", "macros"]
 #! Some features require `alloc` or `std` support and might not work in a `no_std` environment.
 
 ## Enable support for types from the `alloc` crate when running in a `no_std` environment.
-alloc = ["serde_core/alloc", "base64?/alloc", "chrono_0_4?/alloc", "hex?/alloc", "serde_json?/alloc", "time_0_3?/alloc"]
+alloc = ["serde_core/alloc", "base64?/alloc", "bs58?/alloc", "chrono_0_4?/alloc", "hex?/alloc", "serde_json?/alloc", "time_0_3?/alloc"]
```
jonasbb, line 40: Line 62 already enables this unconditionally

jonasbb (changes requested):
I haven't found the time yet to fully review this. I am fine to merge more encoding variants if there is a use for them. I will check for base58, since that is not something I usually see.

One requirement before accepting the PR is to have a test file. You can follow the structure of base64 here too, like https://github.com/jonasbb/serde_with/blob/58eaf03210b6ebda959c468e7b222c30bca8a7f7/serde_with/tests/base64.rs and https://github.com/jonasbb/serde_with/blob/58eaf03210b6ebda959c468e7b222c30bca8a7f7/serde_with/Cargo.toml#L201-L204

serde_with/Cargo.toml
mitinarseny: I guessed there is a convention in this crate to enable `alloc` on all optional dependencies if at least one of them requires it, since this is how it's done for `base64`, too:
```toml
[features]
base58 = ["dep:bs58", "alloc"]
base64 = ["dep:base64", "alloc"]
```
Or do you think we should unconditionally depend on `bs58` with its `alloc` feature enabled by default?

mitinarseny:
@jonasbb Added tests, as you requested. Sorry, didn't notice them earlier.
Also, enriched docs and added impls and tests for `JsonSchemaAs` when one of `schemars` features is enabled.
Does it look better now? 🙂

serde_with/Cargo.toml
jonasbb: This is all good. I got confused here. The alloc on line 62 isn't for the bs58 alloc, but the serde_with alloc.

jonasbb:
Thanks @mitinarseny. This looks good now. Even with schemars test :)
I will prepare a new version this weekend.

jonasbb (approved):

PR #945, opened 2026-05-05
add custom deserializer for String default on null

lms0806:
It currently works like this.
```
#[serde_as]
pub struct applyStruct{
    #[serde_as(deserialize_as = "DefaultOnNull")]
    a: String,
    #[serde_as(deserialize_as = "DefaultOnNull")]
    b: String,
    #[serde_as(deserialize_as = "DefaultOnNull")]
    c: String,
}
```

Through the corresponding PR, the following becomes possible.
```
#[serde_as]
#[apply(
    String => #[serde_as(deserialize_as = "DefaultOnNull")]
)]
pub struct applyStruct{
    a: String,
    b: String,
    c: String,
}
```

### DIFF
--- a/serde_with/tests/serde_as/default_on.rs
+++ b/serde_with/tests/serde_as/default_on.rs
@@ -128,3 +128,24 @@ fn test_default_on_null() {
         expect![[r#"invalid type: integer `3`, expected a string at line 1 column 7"#]],
     );
 }
+
+#[test]
+fn test_default_on_null_with_apply_for_string() {
+    #[serde_with::apply(
+        String => #[serde_as(deserialize_as = "DefaultOnNull")],
+    )]
+    #[serde_with::serde_as]
+    #[derive(Debug, Serialize, Deserialize, PartialEq)]
+    struct S {
+        plain: String,
+        num: u32,
+    }
+
+    check_deserialization(
+        S {
+            plain: String::new(),
+            num: 7,
+        },
+        r#"{"plain":null,"num":7}"#,
+    );
+}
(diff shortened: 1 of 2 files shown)


jonasbb:
Hi thanks for the PR. I am not clear what you want to achieve here though, as there are no extra comments or descriptions in the PR.

> Through the corresponding PR, the following becomes possible.
> 
> ```
> #[serde_as]
> #[apply(
>     String => #[serde_as(deserialize_as = "DefaultOnNull")]
> )]
> pub struct applyStruct{
>     a: String,
>     b: String,
>     c: String,
> }
> ```

In this example the order of the macros is wrong. The order matters and here `apply` must come before `serde_with`.
You did it correctly in the test case.
https://github.com/jonasbb/serde_with/pull/945/changes#diff-1e73a5b8bd968a21b065cb28ab8c092bbe7ff6e6813882e7cbe9e2f568440f9dR134-R137

---

The test `test_default_on_null_with_apply_for_string` already works even without the changes in `serde_with_macros/src/apply.rs`.

I don't want these relaxations in the matching logic:
https://github.com/jonasbb/serde_with/pull/945/changes#diff-318318cff9088da301ddeefbedb9cfeeeb7e6b93ae845571b91cdb5a93b104d3R302-R307
`String` and `std::string::String` can reasonably mean different types. Since the macros only have access to the tokens, I want this to be a bit restrictive in matching, such that you use the same type spelling in the `apply` macro as in your types.

With that I don't think there is something left of this PR that needs merging.

lms0806:
Upon checking based on what you mentioned, I found that I was declaring it incorrectly.

It should have been declared like this, but I was using it like this:

```
#[apply(
String => #[serde_as(deserialize_as = "DefaultOnNull")],

)]
#[serde_with::serde_as]
#[derive(Serialize, Deserialize, Clone, Debug)]
```

I was using it like this:

```
#[serde_with::serde_as]
#[apply(
String => #[serde_as(deserialize_as = "DefaultOnNull")],

)]
#[derive(Serialize, Deserialize, Clone, Debug)]
```

I apologize.

I will close the relevant pull request.

PR #949, opened 2026-05-08
feat: automatic `schemars` support for `Base64`

mitinarseny:
Hey @jonasbb, here is a follow-up PR that adds `schemars` support for `#[serde_as(as = "Base64")]`.
Without it, devs have to annotate it every time:
```rust
#[serde_as]
#[derive(Serialize, Deserialize, JsonSchema)]
struct B64 {
    #[serde_as(as = "Base64")]
    #[schemars(with = "String")] // <- not needed anymore
    bytes: Vec<u8>,
}
```
Will be happy to address any comments and hope it can get included in next release as well 🤞🏻


jonasbb (approved):
Thank you, that looks good.

PR #958, opened 2026-05-21
Add serde_as deserialize_as explain

### LINKED ISSUE #766: docs: include examples of using `#[serde_as(deserialize_as = .. )]` and its counterpart
Hello.

Thanks for making this crate. I generally use it as a convenience for simple things but had to look a little deep to determine if the `#[serde_as]` macro accepted deserialization and serialization specific conversion.

On [the docs](https://docs.rs/serde_with/latest/serde_with/attr.serde_as.html), it is mentioned very briefly, under a wall of text. It may be helpful to newer users if they just see it in the example at the top of the page. Obviously, people should read the docs... but that is not always the case.

I was thinking something like below. Just a visual signpost to say "You can do this".
People familiar with serde and the attributes could intuit them working. For new users to serde and Rust in general, maybe not.

Maybe there is a good place to throw `no_default` on there as well. Is this considered worthwhile for an addition through a PR?

```rust
#[serde_as]
struct Data {
    /// Serialize into number
    #[serde_as(as = "_")]
    a: u32,

    /// Serialize into String,
    /// Deserialize from Bytes or String,
    #[serde_as(
        serialize_as = "DisplayFromStr",
        deserialize_as = "BytesOrString"
    )]
    b: u32,

    /// Serialize into a map from String to String
    #[serde_as(as = "Map<DisplayFromStr, _>")]
    c: Vec<(u32, String)>,

}
```

lms0806:
Added explanation details regarding deserialize_as.

Close #766

### DIFF
--- a/serde_with_macros/src/lib.rs
+++ b/serde_with_macros/src/lib.rs
@@ -456,7 +456,7 @@ fn field_has_attribute(field: &Field, namespace: &str, name: &str) -> bool {
 /// # Example
 ///
 /// ```rust,ignore
-/// use serde_with::{serde_as, DisplayFromStr, Map};
+/// use serde_with::{serde_as, BytesOrString, DisplayFromStr, Map};
 ///
 /// #[serde_as]
 /// #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


jonasbb (commented):
Thank you, this looks good and should hopefully address the discovery.

PR #962, opened 2026-05-28
Modify code after uncommenting unused_qualifications

lms0806:
I uncommented unused_qualifications and proceeded with modifying the code accordingly.


jonasbb (approved):
Thank you for the PR. This looks good now.

PR #968, opened 2026-06-10
Change function position within impl

lms0806:
The modification changed the function position within impl to be the same as the characteristic.


jonasbb:
Hi @lms0806, is there any way to add a test/check that all future implementations are always sorted correctly too?
I'm gonna merge this as it is a neat cleanup.

lms0806:
> Hi @lms0806, is there any way to add a test/check that all future implementations are always sorted correctly too? I'm gonna merge this as it is a neat cleanup.

After looking into it, it seems there is no way to check for this using `cargo check` or `cargo clippy`.

I identified and fixed the issue using the RustRover IDE.

PR #985, opened 2026-08-02
Add support for jiff v0.2

### LINKED ISSUE #936: Feat. req: optionally support `jiff` types
[`jiff`](https://github.com/BurntSushi/jiff) is a datetime library built as an alternative to `chrono`. It would be good if the `Duration*` and `Timestamp*` types can also work here.

willbuckner:
As `chrono` is being  soft-deprecated (see chronotope/chrono#1768; [see also](https://dirkjan.ochtman.nl/writing/2026/01/09/reviewing-2025.html)) and `jiff` is the recommended migration path, it makes sense to add `jiff` support to `serde_with`. Lots of teams are migrating away from `chrono`, and `serde_with` support ends up causing friction for some teams.

Closes #936

### DIFF
--- a/serde_with/src/jiff_0_2.rs
+++ b/serde_with/src/jiff_0_2.rs
@@ -1,0 +1,621 @@
+//! De/Serialization of [jiff v0.2][jiff] types
+//!
+//! This modules is only available if using the `jiff_0_2` feature of the crate.
+//! No extra types are exposed. Instead it enables support for [`jiff_0_2::SignedDuration`] together with [`DurationSeconds`] and its variants.
+//! The types [`jiff_0_2::Timestamp`], [`jiff_0_2::Zoned`], and [`jiff_0_2::civil::DateTime`][::jiff_0_2::civil::DateTime] are supported by [`TimestampSeconds`] and its variants.
+//!
+//! [jiff]: https://docs.rs/jiff/0.2/
+
+// Serialization of large numbers can result in overflows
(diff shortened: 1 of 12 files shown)


serde_with/src/jiff_0_2.rs
```diff
@@ -0,0 +1,621 @@
...
+            Sign::Negative
+        } else {
+            Sign::Positive
+        },
+        std_dur,
+    )
+}
+
+/// Convert a [`DurationSigned`] into a [`SignedDuration`]
+fn duration_from_duration_signed<'de, D>(sdur: DurationSigned) -> Result<SignedDuration, D::Error>
```
jonasbb, line 47: Nit: This could be down below with the other deserialize functions that take a `DurationSigned`.

jonasbb:
Hi, thank you very much for this PR. This addition is very much welcome. I checked the implementation and test. They all look good. Everything is in the same shape as for the other time crates. Even the guide and the schemars support is added :)

willbuckner:
Thanks @jonasbb! Happy to raise another PR to move this function down below if you'd prefer :)

jonasbb:
@willbuckner Thanks for the offer, but thats just a minor thing. It's not necessary.

avsaase:
Sorry to bother you but could you make a release with this change? I'm currently migrating a large codebase to `jiff` and this PR would make that a bit easier.

Issue #986, opened 2026-08-02
`#[skip_serializing_empty]`


akiramusic000:
`#[skip_serializing_empty]` would be a macro similar to `#[skip_serializing_none]`, but instead of applying `#[serde(skip_serializing_if = "Option::is_none")]` to `Option<T>` types, it would apply `#[serde(skip_serializing_if = "Vec::is_empty")]` to `Vec<T>` types.

# Potential problems
- There are other types that have `is_empty` functions, so the name might need to be adjusted, or the implementation adjusted to account for this. However, adjusting the implementation would add complexity that might be unneeded.

jonasbb:
Hi thanks for the proposal and the PR. However, I am not inclined to add a `skip_serializing_empty` macro to this crate.

I suggest you try to use [`serde_with::apply`](https://docs.rs/serde_with/latest/serde_with/attr.apply.html).
Something like this should work:
```rust
#[serde_with::apply(
    Vec => #[serde(skip_serializing_if = "Vec::is_empty")],
)]
```

There is no good generic way to check for emptiness, such as a `IsEmpty` trait. For PR #987 specifically, I think only supporting `Vec` is too specific since std has so many more types.

akiramusic000:
I didn't think about using apply, thanks!

PR #987, opened 2026-08-02
Implement `skip_serializing_empty`

akiramusic000:
First draft of the implementation of `skip_serializing_empty` (See #986). The implementation reuses the existing `skip_serializing_none_add_attr_to_field` (renamed to `skip_serializing_if_add_attr_to_field`). This approach is concise, but very rough around the edges, and probably will need improvements. As well as the rough implementation, I'm not sure how the check that serialize_always is only on `Option` fields could be updated, since `#[skip_serializing_empty]` and `#[skip_serializing_none]` aren't mutually exclusive. In addition to that, the documentation for the macro currently just refers back to the documentation for `skip_serializing_none`.

In summary:
- This implementation is rough and may need to be reworked.
- The check that `serialize_always` is only applied to `Option` will need to be reworked.
- The documentation may need to be updated.
- Tests have not yet been written.

### DIFF
--- a/serde_with_macros/src/lib.rs
+++ b/serde_with_macros/src/lib.rs
@@ -304,15 +304,41 @@ where
 /// ```
 #[proc_macro_attribute]
 pub fn skip_serializing_none(_args: TokenStream, input: TokenStream) -> TokenStream {
-    let res =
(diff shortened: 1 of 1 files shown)


jonasbb:
Please see my response in #986 for the details of the rejection.

